PR #52, opened 2016-06-22
Adapt to vsc-install

stdweird:


bin/mympisanity.py
```diff
@@ -135,15 +139,20 @@ def check():
 if __name__ == '__main__':
     log = getLogger('mympisanity')
     setLogLevelInfo()
+
+    if MPI4PY_EXCEPTION:
+        log.error("No mpi4py found: %s", MPI4PY_EXCEPTION)
+        sys.exit(1)
```
wpoely86: didn't we switch to raising exception upon fatal errors?

stdweird: this is the main of a script, why would you want to raise something here?

lib/vsc/mympirun/rm/__init__.py
```diff
@@ -1,27 +1,30 @@
...
-##
+# along with vsc-mympirun.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Allow other packages to extend this namespace, zip safe setuptools style
+"""
+import pkg_resources
+pkg_resources.declare_namespace(__name__)
 
-import local, pbs
```
wpoely86: shouldn't these be moved?

stdweird: this will probably break the current code, yes. but the `__init__` should be empty, it's up to jeroen to fix this

stdweird:
@JensTimmerman this is now with subclassable shared_setup
@wpoely86 @lepoev: requires new vsc-install in pypi

JensTimmerman:
why do you need to override this vsc.setup inner class, 
can't you just pass

```
setup(
    ...
      cmdclass={'install': mympirun_vsc_install_scripts }
     ...
)
```
(or in the PACKAGE) ?

or does this not work with the overridden inner class things?

JensTimmerman:
oh, it says it in the comments, ` +# because easy_install ignores the easy_install cmdclass `
ok then :+1:

stdweird:
as `vsc_setup` is a class, it seems odd not to do modifications via subclassing. the main reason the original mokey patch code failed is that the entry in the SHARED_SETUP cmdclass was not updated. the `_fvs` trickery in `parse_target` does this now.

stdweird:
test this please

stdweird:
@Iepoev: cherry-pick https://github.com/hpcugent/vsc-mympirun/pull/52/commits/94eb27cd764facaee548fde13bce55760e2cc684 and do a `easy_install --user vsc-install`, should be fixed now

JensTimmerman:
`oh, # because easy_install ignores the easy_install cmdclass`
wel, actually it seems it doesn't have to ignore this anymore, but we choose to delete these unknown command classes in vsc-install, not easy_install ignoring this, vsc-install.

PR #64, opened 2016-07-25
lots of cleanup

Iepoev:
mirror of #53 but to development branch instead of master

PR #65, opened 2016-07-25
Tests

Iepoev:
based on #64 

mirror of PR #58 but to dev branch instead of masterbranch


Iepoev:
this is far from finished

Issue #48, opened 2016-06-14
Don't buffer stdout/stderr


wpoely86:
Currently, mympirun buffers is stdout/stderr and dumps it all at the end. We should stream it.

Caylo:
fixed in #90.

PR #66, opened 2016-07-25
Stream stderr

Iepoev:
based on #64 

Adds an option to stream stderr to a file. This way it is possible to monitor stderr while a job is running instead of having to wait for the entire job to finish.


Iepoev:
this closes #48

Iepoev:
rebased on develop

Iepoev:
this PR should be merged into develop

PR #67, opened 2016-07-26
Stream stderr

Iepoev:
based on #64

Adds an option to stream stderr to a file. This way it is possible to monitor stderr while a job is running instead of having to wait for the entire job to finish.

### DIFF
--- a/lib/vsc/mympirun/option.py
+++ b/lib/vsc/mympirun/option.py
@@ -25,9 +25,11 @@
 """
 Optionparser for mympirun
 """
+import os
 
 from vsc.mympirun.mpi.mpi import MPI
 from vsc.utils.generaloption import GeneralOption
+from vsc.utils import fancylogger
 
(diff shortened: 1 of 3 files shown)


Iepoev:
made a pull request to add the required class to vsc-base/run
https://github.com/hpcugent/vsc-base/pull/234

lib/vsc/mympirun/option.py
```diff
@@ -82,7 +83,11 @@ def make_init(self):
             "double": ("Run double the amount of processes (eg for GAMESS; to change multiplier, use --hybrid)",
                        None, "store_true", False),
 
-            "output": ("filename to write stdout/stderr directly to (instead of stdout)", "str", "store", None),
+            "output": ("redirect the output of mpirun to a file (instead of stdout/stderr)",
```
wpoely86: Can you add a log.info prior to execution of mpirun that repeats to where it will output?

PR #68, opened 2016-07-27
Add tests

Iepoev:


Iepoev:
the tests are failing because jenkins is testing the current masterbranch, while some tests are based on code in the develop branch

stdweird:
jenkins should be testing this branch merged in master, so if this is based on develop, should work fine

Iepoev:
http://pastebin.com/dKYAn6QM

these are comparisons between the attributes returned from `dir(mpi_instance)`

`jenkins_attr_set.difference(local_attr_set)` should contain all attributes that exclusively available on the jenkins mpi_instance. It contains `mpiexec_set_global_options`, which was renamed to `set_mpiexec_global_options` in the develop branch.

https://github.com/hpcugent/vsc-mympirun/blob/master/lib/vsc/mympirun/mpi/mpi.py#L981
https://github.com/hpcugent/vsc-mympirun/blob/develop/lib/vsc/mympirun/mpi/mpi.py#L646

stdweird:
test this please

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
-Tests for the vsc.utils.missing module.
+Tests for the vsc.mympirun.mpi.mpi module.
 
-@author: Jens Timmerman (Ghent University)
+@author: Jeroen De Clerck
 """
 import os
 import stat
-from unittest import TestCase
+import unittest
```
wpoely86: alphabetical order

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+import re
+import string
 
+from vsc.utils.run import run_simple
 from vsc.mympirun.factory import getinstance
-from vsc.mympirun.mpi.mpi import MPI
+import vsc.mympirun.mpi.mpi as mpim
+#from vsc.mympirun.mpi.openmpi import OpenMPI
```
wpoely86: throw out if not needed

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+    #######################
+    ## General functions ##
+    #######################
+
+    def test_what_mpi(self):
+        """test if what_mpi returns the correct mpi flavor"""
+        scriptnames = ["ompirun", "mpirun", "impirun", "mympirun"]
+        for scriptname in scriptnames:
+            # if the scriptname is an executable located on this machine
+            if mpim.which(scriptname):
```
wpoely86: why the `which`? We're testing vsc-mympirun, they should be present?

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+                print("what mpi returns: %s, %s, %s" % (returned_scriptname, mpi, found))
+                # if an mpi implementation was found
+                if mpi:
+                    self.assertTrue(mpi in found,
+                                    msg="returned mpi (%s) is not an element of found_mpi (%s)" % (mpi, found))
+                    self.assertTrue(returned_scriptname == scriptname,
+                                    msg="returned scriptname (%s) doesn't match actual scriptname (%s)" %
+                                    (returned_scriptname, scriptname))
+                else:
+                    self.assertTrue(returned_scriptname.endswith("mpirun"),
```
wpoely86: with if `returned_scriptname` is `None`? (when no mpirun can be found).

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+                                    msg="no mpi found, scriptname should be the path to mpirun but is %s" %
+                                    returned_scriptname)
+
+    def test_stripfake(self):
+        """Test if stripfake actually removes the /bin/fake path in $PATH"""
+        print("old path: %s" % os.environ["PATH"])
+        mpim.stripfake()
+        newpath = os.environ["PATH"]
+        self.assertFalse(("bin/%s" % mpim.FAKE_SUBDIRECTORY_NAME) in newpath, msg="the faked dir is still in $PATH")
```
wpoely86: strictly speaking, this is not enough. There could be something else with puts a `/foo/bin/fake` directory in `$PATH`. Check if there is no `mpirun` in the fake path?

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+
+    def test_which(self):
+        """test if which returns a path that corresponds to unix which"""
+
+        testnames = ["python", "java"]
```
wpoely86: don't use `java`. Try `pwd` or something part of coreutils.

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+
+        for scriptname in testnames:
+            mpiwhich = mpim.which(scriptname)
+            exitcode, unixwhich = run_simple("which " + scriptname)
+            if exitcode > 0:
+                raise Exception("Something went wrong while trying to run `which`: %s" % unixwhich)
+
+            self.assertTrue(mpiwhich, msg="mpi which did not return anything, (unix which: %s" % unixwhich)
+            self.assertEqual(mpiwhich + "\n", unixwhich,
```
wpoely86: do a `strip` on `unixwhich` for the newline?

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+            self.fail("mympirun raised an exception while running main()")
+
+        optdict = mpi_instance.options.__dict__
+
+        print("args given to mympirunoption: %s, instance options: %s, " % (optionparser.args, optdict))
+
+        for opt in optionparser.args:
+            self.assertFalse(opt in optdict)
+
+    # def test_is_mpirun_for(self):
```
wpoely86: why commented out?

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+
+    def test_set_netmask(self):
+        """test if netmask matches the layout of an ip adress"""
+        mpi_instance = getinstance(mpim.MPI, Local, MympirunOption())
+        mpi_instance.set_netmask()
+        # matches "IP address / netmask"
+        reg = re.compile(r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}")
+        print("netmask: %s" % mpi_instance.netmask)
+        for substr in string.split(mpi_instance.netmask, sep=":"):
+            self.assertTrue(reg.match(substr), msg="%s does not seem to be a valid address" % substr)
```
wpoely86: as we're using it anyway, also check if IPy: https://pypi.python.org/pypi/IPy/
Currently 500.500.500.500 is also valid :wink:

test/mpi.py
Iepoev: it's easy to add more mpirun aliasses the the scriptname list. If the which() is dropped you'd have to dig in the code and make sure that the scriptname list only contains the ones that are installed on the system.

test/mpi.py
Iepoev: built-in shell commands don't work with `which`

test/mpi.py
Iepoev: I couldn't find a way to correctly load and unload the modules required for this test

test/mpi.py
```diff
@@ -23,36 +23,251 @@
...
+        mpi_instance.select_device()
+        self.assertTrue(mpi_instance.device and mpi_instance.device in mpi_instance.DEVICE_MPIDEVICE_MAP.values(),
+                        msg="%s is not a valid device type, possible values: %s" %
+                        (mpi_instance.device, mpi_instance.DEVICE_MPIDEVICE_MAP.values()))
+        self.assertTrue(mpi_instance.netmasktype and mpi_instance.netmasktype in mpi_instance.NETMASK_TYPE_MAP.values(),
+                        msg="%s is not a valid netmask type, possible values: %s" %
+                        (mpi_instance.netmasktype, mpi_instance.NETMASK_TYPE_MAP.values()))
+
+    def test_make_node_file(self):
+        """test if the nodefile is made and if it contains the same amount of nodas as mpinodes"""
```
wpoely86: as we do local, we can check if get the correct number of cores (==lines in nodefile)?

test/sched.py
```diff
@@ -0,0 +1,112 @@
...
+from vsc.mympirun.option import MympirunOption
+import vsc.mympirun.rm.sched as schedm
+from vsc.mympirun.rm.local import Local
+from vsc.mympirun.rm.pbs import PBS
+from vsc.mympirun.rm.scoop import Scoop
+from vsc.utils.affinity import sched_getaffinity
+
+SCHEDDICT = {
+    "local": Local,
+    #"pbs": PBS, #doesn't work locally
```
wpoely86: why not, it can be mocked easily?

test/sched.py
```diff
@@ -0,0 +1,112 @@
...
+class TestSched(unittest.TestCase):
+    """tests for vsc.mympirun.mpi.sched functions"""
+
+    def test_what_sched(self):
+        """
+        test if what_sched returns a corresponding scheduler"""
+        for key, val in SCHEDDICT.iteritems():
+            sched, found_sched = schedm.what_sched(key)
+            print("key: %s, sched: %s, found_sched_: %s" % (key, sched, found_sched))
+            sched, found_sched = schedm.what_sched(key)
```
wpoely86: why twice?

test/mpi.py
Iepoev: this happens here https://github.com/hpcugent/vsc-mympirun/pull/68/files#diff-43c74ee7807183fa3e11bc5ec84b7621R188

test/mpi.py
wpoely86: `pwd` is a binary? Anyway, take your pick. As long as it's something more commonly available then java :wink:

test/mpi.py
wpoely86: why load them? Just provide mock `mpirun`?

boegel:
merging this in to get `develop` in a state for further development/testing...

PR #71, opened 2016-08-05
Cpu Pinning

Iepoev:


lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -194,6 +185,24 @@ def mpirun_prepare_execution(self):
 
         return super(IntelMPI, self).mpirun_prepare_execution()
 
+    def pinning_override(self):
+        """ pinning """
+
+        override_type = self.pinning_override_type
```
wpoely86: why assign this to a local variable?

Iepoev: probably leftover from before the cleanup

PR #72, opened 2016-08-08
keep track of .mympirun dir size

Iepoev:


Iepoev:
closes #72

Iepoev:
the temporary dir will not be removed if mympirun is interrupted prematurely.

wpoely86:
Did you check that `mympirun` removes the directory when it exits normally?

Iepoev:
Yes, it does

Issue #50, opened 2016-06-22
Generate the mympirun aliases from the subclasses in setup.py


stdweird:

boegel:
fixed in #73

PR #78, opened 2016-10-11
fix installation issues & fix broken tests

boegel:
this was overlooked in #73 when fixing #50

without this, `setup.py install` fails as below unless the path to `/lib` is already in `$PYTHONPATH`

```
INFO: This is (based on) vsc.install.shared_setup 0.10.14
Traceback (most recent call last):
  File "setup.py", line 40, in <module>
    import vsc.mympirun.mpi.mpi as mpim
ImportError: No module named mympirun.mpi.mpi
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,13 +32,14 @@ Setup for mympirun
 import os
 import pkgutil
 import sys
-
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'lib'))
-
 import vsc.install.shared_setup as shared_setup
 from vsc.install.shared_setup import vsc_setup, log, sdw
-import vsc.mympirun.mpi.mpi as mpim
-from vsc.utils.missing import get_subclasses
+
+
+# hardcoded list, to avoid ugly hacks in order to be able to import from vsc.mympirun in setup.py...
+# this list is checked to be synced via a dedicated unit test
+MYMPIRUN_ALIASES = ['ihmpirun', 'ihmpirun', 'impirun', 'impirun', 'm2hmpirun', 'm2mpirun', 'mhmpirun',
+                    'mmpirun', 'ompirun', 'ompirun']
 
 PACKAGE = {
     'install_requires': [
@@ -71,14 +72,6 @@ class mympirun_vsc_install_scripts(vsc_setup.vsc_install_scripts):
         for loader, modulename, _ in pkgutil.walk_packages([os.path.dirname(mpim.__file__)]):
             loader.find_module(modulename).load_module(modulename)
 
-        mympirun_aliases = []
-
-        for mpiclass in get_subclasses(mpim.MPI):
-            mympirun_aliases += mpiclass._mpiscriptname_for
-
-        mympirun_aliases += ['myscoop']
-
-
         log.info("mympirun_vsc_install_scripts %s", mpim)
         # old-style class
         vsc_setup.vsc_install_scripts.run(self)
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -32,6 +32,9 @@
 import os
 import pkgutil
 import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'lib'))
```
wpoely86: wait, why do we have to do this? Isn't vsc-install take care of this?

boegel: apparently not...

maybe this is taken care of in the package that gets uploaded to PyPI using vsc-install, but to install from the repo (or GH source tarball), you need this there...

@stdweird ?

stdweird:
@boegel this shouldn't be needed, there must something else wrong

setup.py
```diff
@@ -32,6 +32,9 @@
...
+
 import vsc.install.shared_setup as shared_setup
 from vsc.install.shared_setup import vsc_setup, log, sdw
 import vsc.mympirun.mpi.mpi as mpim
```
stdweird: this is the issue. why is the code itself needed in setup.py?

boegel: I'm indeed trying to fix this `import`...

It was introduced in #73

stdweird: you have to replace the `subclasses(MPI)` call with a constant list of classes, and a add a unittest to make sure that this constant in setup.py is equal to the `subclasses(MPI)` result.
otherwise, you'll have to do some ugly hacks to make this work.

boegel: ok, I'll look into that

boegel:
@stdweird added test works: `Make sure that list of mympirun aliases included in setup.py is synced ... ok`

Not sure what the problem is with the other tests failing with this, looks weird... May be an issue with resolving the merge conflicts, but I can't spot it in 1-2-3...

```
OptionConflictError: option --impi-daplud/--enable-impi-daplud/--disable-impi-daplud: conflicting option string(s): --impi-daplud, --enable-impi-daplud, --disable-impi-daplud
```

setup.py
```diff
@@ -34,15 +34,19 @@
...
+# this list is checked to be synced via a dedicated unit test
+MYMPIRUN_ALIASES = ['ihmpirun', 'impirun', 'm2hmpirun', 'm2mpirun', 'mhmpirun', 'mmpirun', 'ompirun']
 
 PACKAGE = {
     'install_requires': [
         'vsc-base >= 2.5.0',
         'vsc-install >= 0.10.9', # for modified subclassing
         'IPy',
     ],
+    'name': 'vsc-mympirun',
```
stdweird: remove

setup.py
```diff
@@ -94,7 +90,7 @@ def run(self):
                 os.chdir(rel_script_dir)
 
                 # create symlinks that point to mympirun for all mpirun aliases
-                for sym_name in mympirun_aliases:
```
stdweird: why did you remove the scoop link?

setup.py
boegel: OK, I looked into this more, and it seems like not specifying the package name like I'm doing here will only work if there's a `PKG-INFO` or `.git/config` file is present from which the GitHub URL can be obtained, and thus the package name.

When installing from the tarball pulled from GitHub, neither of these is there...

Doesn't it make sense to just specify the package name here, just like we do with the version? I don't see the problem with that?

PR #79, opened 2016-10-11
hardcode name of 'fake' subdir in setup.py + add test to verify sync with vsc.mympirun.mpi.mpi

boegel:
missed this in #78...


stdweird:
ok, feel free open an issue in vsc-install to also add setup.py to the prospector tests

PR #88, opened 2016-10-19
check for hanging impi

### LINKED ISSUE #74: wrong SSH key permissions cause hang at MPI startup, no errors whatsoever
faulty permissions on SSH (private) keys which prevent logging into to workernodes through `ssh` seem to cause an MPI job to hang at startup, with no error messages being spit out anywhere

it would be nice if some sort of sanity check is performed to make sure that all workernodes are accessible over SSH, before actually starting `mpirun`

Caylo:
(temp) fixes #74


lib/vsc/mympirun/option.py
```diff
@@ -120,6 +120,8 @@ def make_init(self):
 
             "stats": ("Set MPI statistics level", "int", "store", 0),
 
+            "timeout": ("Set timeout to check for output", "int", "store", 0)
```
boegel: use a sensible default here, e.g. 600s (10m), and make it clear in the help message what the unit is (i.e. seconds)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
     return None
 
 
-class MPI(object):
+class RunFileLoopMPI(RunFile, RunLoop):
+    """
+    RunFile specific for MPI because intel doesn't feel like fixing their shit
+    """
+    def __init__(self, cmd, **kwargs):
+        self.options = kwargs.pop('options', None)
```
boegel: add a docstring please

just pass in the value for `output` and `timeout`, rather than *all* options?

```python
self.output = kwargs.pop('output', None)
self.timeout = kwargs.pop('timeout', None)
```

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+        self.counter = 1
+        super(RunFileLoopMPI, self).__init__(cmd, **kwargs)
+        self.readsize = -1
```
boegel: you can drop this?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
 
 
-class MPI(object):
+class RunFileLoopMPI(RunFile, RunLoop):
+    """
+    RunFile specific for MPI because intel doesn't feel like fixing their shit
+    """
+    def __init__(self, cmd, **kwargs):
+        self.options = kwargs.pop('options', None)
+        self.counter = 1
```
boegel: use a more meaningful name, maybe `self.timeout_cnt`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+        super(RunFileLoopMPI, self).__init__(cmd, **kwargs)
+        self.readsize = -1
+        self.seen_output = False
+
+    def _loop_process_output(self, output):
+        """
+        check if process is generating any output at all; if not, warn the user after a set amount of time
+        """
+        del output
```
boegel: because we're (also) deriving from `RunFile`, `output` should always be empty, otherwise something weird is going on...

So, add something like:

```python
if output:
    # raise hell
```

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+        if not self.seen_output:
```
boegel: I would avoid too deep indentation here, and go with:

```python
# early exit in case we've already seen output
if self.seen_output:
    return

...

File: lib/vsc/mympirun/mpi/mpi.py [hunk 7]
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+            self.readfile = open(self.filename, 'r')
```
boegel: not used?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+            with open(self.options.output, 'r') as fin:
+                if len(fin.read())>0:
```
boegel:
```python
if fin.read():
```

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
...
+                    self.seen_output = True
+
+            time_passed = self._loop_count * self.LOOP_TIMEOUT_MAIN
+            if time_passed > self.options.timeout * self.counter and not self.seen_output:
+                print '\n'.join(["WARNING: mympirun has been running for %s seconds without any output." % time_passed,
+                "This means your program may be hanging. Please check if this is normal.",
+                "If not, check your SSH key permissions? (private key should be protected)"])
```
boegel: compose the message first, and print it to `stderr`
I also wouldn't mention the SSH key thing, since it may not be entirely clear to users *what* to check exactly, and there are other possible reasons for a hang

```python
warning_msg = '\n'.join([
    ""WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
    "This may mean that your program is hanging, please check and make sure that is not the case!",
    '',
    "If this warning is printed too soon and the program is doing useful work without producing any output,",
    "you can increase the timeout threshold via --timeout (current setting: %s seconds)" % self.timeout,
])
sys.stderr.write(warning_msg + '\n')
```

lib/vsc/mympirun/mpi/mpi.py
boegel: actually, just grab `filename` , but do pass it down (don't pop it from `kwargs`):

```python
self.timeout = kwargs.pop('timeout', None)
self.filename = kwargs.get('filename', None)
```

lib/vsc/mympirun/option.py
```diff
@@ -120,6 +120,8 @@ def make_init(self):
 
             "stats": ("Set MPI statistics level", "int", "store", 0),
 
+            "timeout": ("Set timeout to check for output (seconds)", "int", "store", 600),
```
boegel: `"Set timeout threshold to print warning if no output has been produced yet"`

test/end2end.py
```diff
@@ -120,3 +120,28 @@ def test_output(self):
...
+        self.assertEqual(ec, 0, "Command exited normally: exit code %s; output: %s" % (ec, out))
+
+        warning = "WARNING: mympirun has been running for %s seconds without any output."
+
+        regex = re.compile(warning % 3)
+        self.assertTrue(regex.search(out), "Pattern '%s' found in: %s" % (regex.pattern, out))
+
+        # should not warn again when there has been some output
+        regex = re.compile(warning % 5)
+        self.assertFalse(regex.search(out), "Pattern '%s' found in: %s" % (regex.pattern, out))
```
boegel: do ` regex.findall` using `""WARNING: mympirun has been running for .* seconds without any output."` as a regex pattern

you find find *exactly* one match, on the 3 second mark

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +151,37 @@ def which(cmd):
     return None
 
 
-class MPI(object):
+class RunFileLoopMPI(RunFile, RunLoop):
+    """
+    RunFile specific for MPI because intel doesn't feel like fixing their shit
```
boegel: yeah, maybe be less aggressive here.. ;)

`
"""
Combination of RunFile and RunLoop to support output to file,
while also checking whether any output has been produced after a specified amount of time.
"""

lib/vsc/mympirun/mpi/mpi.py
boegel: actually.... ;-)

since `timeout` is a *required* argument, go with:

```python
def __init__(self, cmd, output_timeout, *args, **kwargs):
```

lib/vsc/mympirun/option.py
boegel: also `--timeout` is too general, it's more like `--output-check-timeout`

lib/vsc/mympirun/mpi/mpi.py
Caylo: :)

boegel:
Jenkins: test this please

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,50 @@ def which(cmd):
...
+    """
+    Combination of RunFile and RunLoop to support output to file,
+    while also checking whether any output has been produced after a specified amount of time.
+    """
+    def __init__(self, cmd, **kwargs):
+        """
+        handle initialisation: get filename and output timeout from arguments
+        """
+        self.filename = kwargs.get('filename', None)
+        self.output_timeout = kwargs.pop('output_timeout', None)
```
boegel: please make sure the value here makes sense, i.e. can be parsed as an `int`; if not, raise a `ValueError`

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,50 @@ def which(cmd):
...
+        super(RunFileLoopMPI, self).__init__(cmd, **kwargs)
+
+        self.timeout_cnt = 1
```
boegel: shouldn't this be `0`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,50 @@ def which(cmd):
...
+    def _loop_process_output(self, output):
+        """
+        check if process is generating any output at all; if not, warn the user after a set amount of time
+        """
+        if output:
+            print "Output was found using RunFile:\n%s\n This means something went horribly wrong." % output
 
+        if self.seen_output:
+            return
+        else:
```
boegel: since you `return` in the `if`, you don't need to put this in an `else` (less indent)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,50 @@ def which(cmd):
...
+
+        self.timeout_cnt = 1
+        self.seen_output = False
+
+    def _loop_process_output(self, output):
+        """
+        check if process is generating any output at all; if not, warn the user after a set amount of time
+        """
+        if output:
+            print "Output was found using RunFile:\n%s\n This means something went horribly wrong." % output
```
boegel: do a `raise ValueError("...")` here, because if `output` is defined to anything, there's a bug somewhere

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -301,7 +344,7 @@ def main(self):
         # actual execution
         self.log.debug("main: going to execute cmd %s", " ".join(self.mpirun_cmd))
         self.log.info("writing mpirun output to %s", self.options.output)
-        exitcode, _ = run_to_file(self.mpirun_cmd, filename=self.options.output)
+        exitcode, _ = RunFileLoopMPI.run(self.mpirun_cmd, output_timeout=self.options.output_check_timeout, filename=self.options.output)
```
boegel: trim long line

```python
exitcode, _ = RunFileLoopMPI.run(self.mpirun_cmd, output_timeout=self.options.output_check_timeout,
                                 filename=self.options.output)
```

test/end2end.py
```diff
@@ -120,3 +120,22 @@ def test_output(self):
         with open(f_out, 'r') as output:
             text = output.read()
             self.assertTrue(regex.match(text), "Pattern '%s' found in: %s" % (regex.pattern, text))
+
+
+    def test_hanging(self):
+        """ Test --timeout option when program has no output"""
+        no_output_mpirun = '\n'.join([
+            "#!/bin/bash",
+            "sleep 4", # sleep for 4 seconds
```
boegel: isn't the comment a bit of overkill? ;)

test/end2end.py
```diff
@@ -120,3 +120,22 @@ def test_output(self):
...
+            "echo 'some output'",
+            "sleep 3"
+        ])
+
+        install_fake_mpirun('mpirun', self.tmpdir, txt=no_output_mpirun)
+        ec, out = run_simple("%s %s --setmpi impirun --output-check-timeout 2 hostname" % (sys.executable, self.mympiscript))
+        self.assertEqual(ec, 0, "Command exited normally: exit code %s; output: %s" % (ec, out))
+
+        regex = re.compile("WARNING: mympirun has been running for [1-9] seconds without seeing any output.")
```
boegel: use `.*` rather than `[1-9]`, to make sure you catch any wrongly included warnings (e.g. 0 seconds)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,57 @@ def which(cmd):
...
+        handle initialisation: get filename and output timeout from arguments
+        """
+        self.filename = kwargs.get('filename', None)
+        self.output_timeout = kwargs.pop('output_timeout', None)
+
+        # check timeout
+        try:
+            self.output_timeout = int(self.output_timeout)
+        except ValueError:
+            LOGGER.raiseException("Invalid timeout value: %s (should be integer)" % self.output_timeout)
```
boegel: don't hide the actual conversion error:

```python
except ValueError as err:
    raise ValueError(""Invalid timeout value: %s (should be integer): %s" % (self.output_timeout, err))

**Comment by boegel** [comment]
Jenkins: test this please

File: lib/vsc/mympirun/mpi/mpi.py [hunk 21]
```diff
@@ -151,8 +152,57 @@ def which(cmd):
...
+        # check timeout
+        try:
+            self.output_timeout = int(self.output_timeout)
+        except ValueError as err:
+            raise ValueError("Invalid timeout value: %s (should be integer): %s" % (self.output_timeout, err))
+
+        super(RunFileLoopMPI, self).__init__(cmd, **kwargs)
+
+        self.timeout_cnt = 1
+        self.seen_output = self.output_timeout<0 #no check when output_timeout is negative
```
boegel: please reformat to

```python
# no output check when timeout is negative
self.seen_output = self.output_timeout < 0 
```

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,57 @@ def which(cmd):
...
+                    self.seen_output = True
+
+            time_passed = self._loop_count * self.LOOP_TIMEOUT_MAIN
+            if time_passed > self.output_timeout * self.timeout_cnt and not self.seen_output:
+                warning_msg = '\n'.join([
+                "WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
+                "This may mean that your program is hanging, please check and make sure that is not the case!",
+                '',
+                "If this warning is printed too soon and the program is doing useful work without producing any output,",
+                "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)" % self.output_timeout,
```
boegel: line too long (over 120 chars)?

drop the `else` above, move indent to the left, and you're good I think :)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -151,8 +152,57 @@ def which(cmd):
...
+
+        # check timeout
+        try:
+            self.output_timeout = int(self.output_timeout)
+        except ValueError as err:
+            raise ValueError("Invalid timeout value: %s (should be integer): %s" % (self.output_timeout, err))
+
+        super(RunFileLoopMPI, self).__init__(cmd, **kwargs)
+
+        self.timeout_cnt = 1
```
boegel: why not `0`?

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Thanks @Caylo!

Issue #89, opened 2016-10-26
stream stdout instead of dumping from RunFile


Caylo:
Output should be unbuffered. see https://github.com/hpcugent/vsc-mympirun/blob/develop/lib/vsc/mympirun/mpi/mpi.py#L307

boegel:
fixed in #90

PR #90, opened 2016-10-26
Stream stdout

Caylo:
fixing #89

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -150,16 +150,15 @@ def which(cmd):
     LOGGER.warning("Could not find command '%s' (with permissions to read/execute it) in $PATH (%s)", cmd, paths)
     return None
 
-def check_output(time_passed, output_timeout, seen_output):
+def check_output(time_passed, output_timeout):
     """
     Check whether output has been produced after a specified time
 
     @param time_passed: the time passed since starting the program
     @param output_timeout: the treshhold (--output-check-timeout)
-    @param seen_output: true if either output has been observed or the warning msg has been printed
     """
     warning_msg = None
-    if time_passed > output_timeout and not seen_output:
+    if time_passed > output_timeout:
         warning_msg = '\n'.join([
             "WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
             "This may mean that your program is hanging, please check and make sure that is not the case!",
@@ -168,9 +167,8 @@ def check_output(time_passed, output_timeout, seen_output):
             "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)" % output_timeout,
             ])
         LOGGER.warn(warning_msg + '\n')
-        seen_output = True
 
-    return seen_output
+    return warning_msg
 
 class RunFileLoopMPI(RunFile, RunLoop):
     """
@@ -202,7 +200,8 @@ class RunFileLoopMPI(RunFile, RunLoop):
             raise IOError("Couldn't check file size; %s" % err)
 
         time_passed = self.LOOP_TIMEOUT_INIT + self._loop_count * self.LOOP_TIMEOUT_MAIN
-        self.seen_output = check_output(time_passed, self.output_timeout, self.seen_output)
+        if not self.seen_output:
+            self.seen_output = check_output(time_passed, self.output_timeout) is not None
 
 
 class RunAsyncMPI(RunAsyncLoopStdout):
@@ -224,7 +223,8 @@ class RunAsyncMPI(RunAsyncLoopStdout):
             self.seen_output = True
 
         time_passed = self.LOOP_TIMEOUT_INIT + self._loop_count * self.LOOP_TIMEOUT_MAIN
-        self.seen_output = check_output(time_passed, self.output_timeout, self.seen_output)
+        if not self.seen_output:
+            self.seen_output = check_output(time_passed, self.output_timeout) is not None
 
         super(RunAsyncMPI, self)._loop_process_output(output)
 
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -150,6 +150,20 @@ def which(cmd):
     LOGGER.warning("Could not find command '%s' (with permissions to read/execute it) in $PATH (%s)", cmd, paths)
     return None
 
+def check_output(time_passed, output_timeout, seen_output):
+    if time_passed > output_timeout and not seen_output:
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: test this please

boegel:
@JensTimmerman any idea why I need to keep asking Jenkins to test new commits?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -198,6 +198,42 @@ def _loop_process_output(self, output):
             self.log.error(warning_msg + '\n')
             self.seen_output = True
 
+class RunAsyncMPI(RunAsyncLoopStdout):
+    """
+    RunAsync for MPI in case a program 'hangs'
```
boegel: this doesn't really have anything to do with hanging, it's just a matter of producing useful output as soon as it's available, mainly for interactive use...

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -198,6 +198,42 @@ def _loop_process_output(self, output):
...
+
+    """
+    def __init__(self, cmd, **kwargs):
+        self.output_timeout = kwargs.pop('output_timeout', None)
+        # check timeout
+        try:
+            self.output_timeout = int(self.output_timeout)
+        except ValueError as err:
+            raise ValueError("Invalid timeout value: %s (should be integer): %s" % (self.output_timeout, err))
```
boegel: please do the conversion in `main` rather than duplicating the code

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -198,6 +198,42 @@ def _loop_process_output(self, output):
...
+        if time_passed > self.output_timeout and not self.seen_output:
+            warning_msg = '\n'.join([
+            "WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
+            "This may mean that your program is hanging, please check and make sure that is not the case!",
+            '',
+            "If this warning is printed too soon and the program is doing useful work without producing any output,",
+            "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)" % self.output_timeout,
+            ])
+            self.log.error(warning_msg + '\n')
+            self.seen_output = True
```
boegel: same here, don't copy-paste this, hoist it into a function you call to that produces the warning message (but don't pass down `self.log`, do that locally)

boegel:
@JensTimmerman I updated the list of approved people in the `vsc-mympirun` PR builder config on Jenkins to include @Caylo, that may be sufficient, we'll know on the next push...

test/end2end.py
```diff
@@ -139,3 +139,23 @@ def test_hanging(self):
...
+            "#!/bin/bash",
+            "sleep 4",
+            "echo 'some output'",
+            "sleep 3"
+        ])
+
+        f_out = os.path.join(self.tmpdir, "temp.out")
+
+        install_fake_mpirun('mpirun', self.tmpdir, txt=no_output_mpirun)
+        ec, out = run_simple("%s %s --setmpi impirun --output %s --output-check-timeout 2 hostname" % (sys.executable,
```
boegel: avoid wrapping using

```python
cmd = "%s %s ..." sys.executable,
  
            
  Write  Preview

avoid wrapping using

```python
cmd = "%s %s ..." % (...)
ec, out = run_simple(cmd)

**Comment by JensTimmerman** [comment]
@boegel you need to tell jenkins this pr is whitelisted before it will auto test it, otherwise it will ask on every commit: jenkins add to whitelist

File: lib/vsc/mympirun/mpi/mpi.py [hunk 5]
```diff
@@ -150,6 +150,20 @@ def which(cmd):
     LOGGER.warning("Could not find command '%s' (with permissions to read/execute it) in $PATH (%s)", cmd, paths)
     return None
 
+def check_output(time_passed, output_timeout, seen_output):
+    if time_passed > output_timeout and not seen_output:
```
boegel: add a docstring please (and document the function arguments)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -150,6 +150,20 @@ def which(cmd):
...
+            "WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
+            "This may mean that your program is hanging, please check and make sure that is not the case!",
+            '',
+            "If this warning is printed too soon and the program is doing useful work without producing any output,",
+            "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)" % output_timeout,
+            ])
+        LOGGER.warn(warning_msg + '\n')
+        seen_output = True
+
+    return seen_output
```
boegel: why not return the actual message itself, or `None` if no message should be printed?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -340,19 +363,22 @@ def main(self):
...
-        if print_output:
-            with open(self.options.output, 'r') as fin:
-                print(fin.read())
+
+        # check timeout
+        try:
+            self.options.output_check_timeout = int(self.options.output_check_timeout)
+        except ValueError as err:
+            raise ValueError("Invalid timeout value: %s (should be integer): %s" % (self.options.output_check_timeout,
+                             err))
```
boegel: try to avoid wrapping (shorten the error message, e.g. use `int`)

lib/vsc/mympirun/mpi/mpi.py
Caylo: if seen_output was already `True`, it should be returned as it was to avoid printing the warning message multiple times

boegel: but you can check that where you're calling this function, no?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -187,17 +202,32 @@ def _loop_process_output(self, output):
...
+
+        self.seen_output = self.output_timeout < 0 #no check when output_timeout is negative
+
+    def _loop_process_output(self, output):
+        """ Send output to stdout + hang check """
+        if len(output) > 0:
             self.seen_output = True
 
+        time_passed = self.LOOP_TIMEOUT_INIT + self._loop_count * self.LOOP_TIMEOUT_MAIN
+        self.seen_output = check_output(time_passed, self.output_timeout, self.seen_output)
```
boegel: @Caylo passing `self.seen_output` and then updating it is silly... the `seen_output` logic should be handled here, not in the `check_output` function

boegel:
Thanks @Caylo!

PR #91, opened 2016-11-09
output-check-fatal option (true by default)

Caylo:
wrt the remarks in #74

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -61,10 +61,10 @@ LOGGER = getLogger()
 
 TIMEOUT_CODE = 124
 TIMEOUT_WARNING = """mympirun has been running for %s seconds without seeing any output.
-                  This may mean that your program is hanging, please check and make sure that is not the case!
+This may mean that your program is hanging, please check and make sure that is not the case!
 
-                  If this warning is printed too soon and the program is doing useful work without producing any output,
-                  you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"""
+If this warning is printed too soon and the program is doing useful work without producing any output,
+you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"""
 
 TIMEOUT_FATAL_MSG = "This is considered fatal (unless --disable-output-check-fatal is used)"
 
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -45,7 +45,7 @@ from IPy import IP
 
 from vsc.utils.fancylogger import getLogger
 from vsc.utils.missing import get_subclasses, nub
-from vsc.utils.run import RunAsyncLoopStdout, RunFile, RunLoop, run_simple
+from vsc.utils.run import Run, RunAsyncLoopStdout, RunFile, RunLoop, run_simple
 
 # part of the directory that contains the installed fakes
 INSTALLATION_SUBDIRECTORY_NAME = '(VSC-tools|(?:vsc-)?mympirun)'
@@ -60,12 +60,12 @@ TEMPDIR_ERROR_SIZE = 1000000
 LOGGER = getLogger()
 
 TIMEOUT_CODE = 124
-TIMEOUT_WARNING = '\n'.join([
-    "mympirun has been running for %s seconds without seeing any output.",
-    "This may mean that your program is hanging, please check and make sure that is not the case!",
-    '',
-    "If this warning is printed too soon and the program is doing useful work without producing any output,",
-    "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"])
+TIMEOUT_WARNING = """mympirun has been running for %s seconds without seeing any output.
+                  This may mean that your program is hanging, please check and make sure that is not the case!
+
+                  If this warning is printed too soon and the program is doing useful work without producing any output,
+                  you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"""
+
 TIMEOUT_FATAL_MSG = "This is considered fatal (unless --disable-output-check-fatal is used)"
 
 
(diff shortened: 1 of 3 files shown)


lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -202,9 +214,13 @@ def _loop_process_output(self, output):
         if not self.seen_output:
             msg = check_output(time_passed, self.output_timeout)
             if msg:
-                self.log.warn(msg)
                 # avoid getting warning multiple times by setting seen_output to True if a warning was produced
                 self.seen_output = True
+                if self.fatal_option:
+                    self.stop_tasks()
+                    raise RunMPIException(TIMEOUT_EXIT_CODE, fatal_msg(msg))
```
boegel: I would just use `self.log.error` and exit (raising an error means trowing a traceback showing the guts of the code, which is not useful to users).
Also, always do the `self.log.warn` with the standard message, and just emit an additional error message in case fatal is enabled:

```python
self.log.warn(msg)
if self.fatal_option:
    self.stop_tasks()
    error_msg = "No output detected after %d seconds" % ...
    error_msg += "; this is considered fatal (unless --disable-output-check-fatal is used)"
    self.log.error(error_msg)
    sys.exit(TIMEOUT_EXIT_CODE)
```

boegel: since you need this in two places, use a function to get both the error message and exit code

```python
error_msg, exit_code = output_timeout_error()
self.log.error(error_msg)
sys.exit(exit_code)
```

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -229,13 +246,27 @@ def _loop_process_output(self, output):
...
+                else:
+                    self.log.warn(msg)
 
         super(RunAsyncMPI, self)._loop_process_output(output)
 
 
+class RunMPIException(Exception):
```
boegel: trash this, just use `self.log.error` and `sys.exit`

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -389,9 +420,10 @@ def main(self):
 
         if self.options.output:
             exitcode, _ = RunFileLoopMPI.run(self.mpirun_cmd, output_timeout=self.options.output_check_timeout,
-                                             filename=self.options.output)
+                                             filename=self.options.output, fatal_option=self.options.output_check_fatal)
         else:
-            exitcode, _ = RunAsyncMPI.run(self.mpirun_cmd, output_timeout=self.options.output_check_timeout)
+            exitcode, _ = RunAsyncMPI.run(self.mpirun_cmd, output_timeout=self.options.output_check_timeout,
+                                          fatal_option=self.options.output_check_fatal)
```
boegel: there's lot of common stuff here, so how about something like:

```python
run_kwargs = {
    'fatal_option': self.options.output_check_fatal,
    'output_timeout': self.options.output_check_timeout,
}
if self.options.output:
    run_mpirun_cmd = RunFileLoopMPI.run
    run_kwargs.update({'filename': self.options.output})
else:
    run_mpirun_cmd = RunAsyncMPI.run
exitcode, _ = run_mpirun_cmd(self.mpirun_cmd, **run_kwargs)
```

but do come up with a better name than `fatal_option`... ;-)

lib/vsc/mympirun/option.py
```diff
@@ -94,7 +94,9 @@ def make_init(self):
                        "str", "store", None),
 
             "output-check-timeout": ("Set timeout threshold to print warning if no output has been produced yet (sec)",
-                                     "int", "store", 600),
+                                     "int", "store", 3600),
+
+            "output-check-fatal": ("Set output check to be fatal after a set amount of time", None, "store_true", True),
```
boegel: `"Exit with an error in case of output check timeout"`

test/end2end.py
```diff
@@ -166,3 +175,33 @@ def test_hanging_file(self):
...
+            "WARNING: mympirun has been running for .* seconds without seeing any output.",
+            "If you don't wish for this check to stop your program, run with --disable-output-check-fatal.",
+        ]
+
+        for regex in regexes:
+            regex = re.compile(regex)
+            self.assertTrue(len(regex.findall(out)) == 1, "Pattern '%s' found in: %s" % (regex.pattern, out))
+
+
+
```
boegel: more whitespace, more! ;)

stdweird:
i'm not a big fan of the "true by default", as it will fail previously working scripts for no obvious reason.

boegel:
@stdweird I feel the potential benefits (i.e. potentially saving several years of compute time from being wasted...) outweigh the potential annoyance when we start using this in production.
The error message this produces is very clear.

Also, if this turns out to be too intrusive, we can a) disable it by default (which probably means nobody will ever use it), b) bump the timeout threshold (now set to 1 hour, in this PR).

No output whatsoever being produced after 1 hour is very suspicious, if you ask me, I don't think I've software that behaves like that (other than trivial toy examples maybe, and even then...).

lib/vsc/mympirun/option.py
```diff
@@ -94,7 +94,9 @@ def make_init(self):
                        "str", "store", None),
 
             "output-check-timeout": ("Set timeout threshold to print warning if no output has been produced yet (sec)",
```
stdweird: you should also clarify here that `output` means `stdout` (and/or `stderr`), not any other form of output (like result files).
maybe rephrase like `Warn when no stdout/stderr was seen after start (in seconds; negative number disables this test)`?

lib/vsc/mympirun/option.py
stdweird: or `Exit with exitcode %s instead of warn if --output-check-timout is enabled` (well, `--output-check-timeout` is always enabled, but when non-negative).

stdweird:
@boegel ok. with an hour, thinigs should be safer. and the message is indeed very clear.

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -58,7 +59,14 @@
...
 LOGGER = getLogger()
 
-TIMEOUT_EXIT_CODE = 124
+TIMEOUT_CODE = 124
+TIMEOUT_WARNING = '\n'.join([
+    "mympirun has been running for %s seconds without seeing any output.",
+    "This may mean that your program is hanging, please check and make sure that is not the case!",
+    '',
+    "If this warning is printed too soon and the program is doing useful work without producing any output,",
+    "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"])
```
boegel: don't use `.join` here, it's static data, use `"""..."""` for multi-line strings

```python
TIMEOUT_WARNING = """mympirun has been ...
...
seconds)"""

File: lib/vsc/mympirun/mpi/mpi.py [hunk 8]
```diff
@@ -150,24 +159,21 @@ def which(cmd):
     LOGGER.warning("Could not find command '%s' (with permissions to read/execute it) in $PATH (%s)", cmd, paths)
     return None
 
-def check_output(time_passed, output_timeout):
+def loop_process_output_common(runmpiclass):
```
boegel: create a separate subclass of `Run` for this, which adds this method, and then derive `RunFileLoopMPI` and `RunAsyncMPI` (also) from that subclass (so the method is available)

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -150,24 +159,21 @@ def which(cmd):
...
-        warning_msg = '\n'.join([
-            "WARNING: mympirun has been running for %s seconds without seeing any output." % time_passed,
-            "This may mean that your program is hanging, please check and make sure that is not the case!",
-            '',
-            "If this warning is printed too soon and the program is doing useful work without producing any output,",
-            "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)" % output_timeout,
-            ])
-
-    return warning_msg
+    time_passed = runmpiclass.LOOP_TIMEOUT_INIT + runmpiclass._loop_count * runmpiclass.LOOP_TIMEOUT_MAIN
```
boegel: change this back to `self.` in the method of the proper `Run` subclass

lib/vsc/mympirun/option.py
```diff
@@ -93,8 +93,11 @@ def make_init(self):
             "output": ("redirect the output of mpirun to a file (instead of stdout/stderr)",
                        "str", "store", None),
 
-            "output-check-timeout": ("Set timeout threshold to print warning if no output has been produced yet (sec)",
-                                     "int", "store", 600),
+            "output-check-timeout": ("Warn when no stdout/stderr was seen after start (in seconds; negative number "
+                                     "disables this test", "int", "store", 3600),
```
boegel: @Caylo let's introduce a constant for `3600` too?

test/end2end.py
```diff
@@ -158,11 +166,33 @@ def test_hanging_file(self):
...
+    def test_hanging_fatal(self):
+        """Test fatal hanging check when program has no output"""
+        no_output_mpirun = '\n'.join([
+            "#!/bin/bash",
+            "sleep 4",
+        ])
+
+        install_fake_mpirun('mpirun', self.tmpdir, txt=no_output_mpirun)
+        cmd = ' '.join([
```
boegel: add a comment here and mention that `--output-check-fatal` is enabled by default

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -60,12 +60,12 @@
...
 
 TIMEOUT_CODE = 124
-TIMEOUT_WARNING = '\n'.join([
-    "mympirun has been running for %s seconds without seeing any output.",
-    "This may mean that your program is hanging, please check and make sure that is not the case!",
-    '',
-    "If this warning is printed too soon and the program is doing useful work without producing any output,",
-    "you can increase the timeout threshold via --output-check-timeout (current setting: %s seconds)"])
+TIMEOUT_WARNING = """mympirun has been running for %s seconds without seeing any output.
+                  This may mean that your program is hanging, please check and make sure that is not the case!
```
boegel: @Caylo if you indent like this, the indentation is also included in the printed warning?

test/end2end.py
```diff
@@ -184,6 +184,7 @@ def test_hanging_fatal(self):
         ])
 
         install_fake_mpirun('mpirun', self.tmpdir, txt=no_output_mpirun)
+        # fatal is True by default
```
boegel: @Caylo please mention that actual CLI option

boegel:
Thanks @Caylo!

PR #92, opened 2016-11-16
define variables outside of try/except

Caylo:


boegel:
@Caylo let's catch this issue with a test...

boegel:
lgtm, thanks @Caylo!

Issue #34, opened 2014-04-24
mympirun is reported to only start 1 process on each physical node


ehiggs:
Users are reporting that mympirun is starting only 1 process per physical node. So in a job where you have nodes=4:ppn=2 and 2 of these logical nodes land on the same physical node, then it's not starting multiple processes. If this is the case, it's really broken.

Caylo:
@stdweird is this still an issue?

stdweird:
@Caylo dunno, maybe just try it?

boegel:
@Caylo some debug info using current `develop` (after merging of #92) @ https://github.ugent.be/gist/kehoste/b484c8435dc07dbce50a65694e853687

boegel:
@Caylo in that example, Torque gave us 16 cores on one node, 4 cores on the other node, so `mympirun` should be honouring that:

* in 'normal' mode, start 16 processes on 1st node, 4 processes on other node
* with `--hybrid 2`, start 2 processes on each node (`#hybrid` per unique node), but also print a warning (only for hybrid mode) w.r.t. the imbalance of cores-per-node (combining `--hybrid` and partial nodes doesn't make much sense)
* with `--double`, start 32 processes on 1st node, 8 processes on other node

boegel:
fixed in #93

PR #93, opened 2016-11-16
fix logic in scheduler

Caylo:
ref #34 
todo: check if we need `mpiprocesspernode` var at all?

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -575,18 +575,21 @@ class MPI(object):
 
     def get_universe_ncpus(self):
         """Construct dictionary with number of processes to start per node, based on --universe"""
-        universe_ppn = {}
-        for node in nub(self.nodes):
-            universe_ppn[node] = 0
-        i = 0
-        while i < self.options.universe:
-            for node, ppn in universe_ppn.items():
-                if ppn < self.ppn_dict[node]:
-                    universe_ppn[node] = ppn+1
-                    i += 1
-                if i >= self.options.universe:
-                    break
-
+        if self.options.universe > len(self.nodes):
+            ex = "Universe asks for more processes (%s) than available processors (%s)"
+            self.log.raiseException(ex % (self.options.universe, len(self.nodes)))
+        nodes = nub(self.nodes)
+        universe_ppn = dict((node, 0) for node in nodes)
+        proc_cnt = 0
+        node = nodes.pop(0)
+        while proc_cnt < self.options.universe:
+            if universe_ppn[node] < self.ppn_dict[node]:
+                universe_ppn[node] += 1
+                proc_cnt += 1
+                # recycle node
+                nodes.append(node)
+            # select next node to assign a process to
+            node = nodes.pop(0)
         return universe_ppn
 
 
@@ -809,7 +812,7 @@ class MPI(object):
         """Hydra specific mpiexec options."""
         self.get_hydra_info()
         # see https://software.intel.com/en-us/articles/controlling-process-placement-with-the-intel-mpi-library
-        # machinefile keeps the imbalance if there is one; hostfile doesn't
+        # --machinefile keeps the imbalance if there is one; --hostfile doesn't
         self.mpiexec_options.append("--machinefile %s" % self.mpiexec_node_filename)
         if self.options.branchcount is not None:
             self.mpiexec_options.append("--branch-count %d" % self.options.branchcount)
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -27,6 +27,7 @@ Base MPI class, all actual classes should inherit from this one
 
 @author: Stijn De Weirdt
 @author: Jeroen De Clerck
+@author: Caroline De Brouwer
 """
 
 import os
@@ -414,8 +415,8 @@ class MPI(object):
     def check_usable_cpus(self):
         """Check and log if non-standard cpus (eg due to cpusets)."""
         if not self.cores_per_node == len(self.cpus):
-            self.log.info("check_usable_cpus: non-standard cpus found: requested ppn %s, found cpus %s, usable cpus %s",
-                          self.ppn, self.cores_per_node, len(self.cpus))
+            self.log.info("check_usable_cpus: non-standard cpus found: found cpus %s, usable cpus %s",
+                          self.cores_per_node, len(self.cpus))
 
     def check_limit(self):
         """Check if the softlimit of the stack exceeds 1MB, if it doesn't, show an error."""
@@ -428,7 +429,7 @@ class MPI(object):
         """
         Sets ompthreads to the amount of threads every MPI process should use.
 
-        For example, with hybrid 2 every MPI process should have a total 2 threads (each on a seperate processors).
+        For example, with hybrid 2 every MPI process should have a total 2 threads (each on a seperate processor).
         This way each node will have 8 MPI processes (assuming ppn is 16). Will default to 1 if hybrid is disabled.
         """
         if 'OMP_NUM_THREADS' in os.environ:
@@ -573,7 +574,7 @@ class MPI(object):
 
     def get_universe_ncpus(self):
         """Return ppn for universe"""
-        return self.mpiprocesspernode
+        return self.options.universe
 
     def make_mympirundir(self):
         """
@@ -781,7 +782,7 @@ class MPI(object):
         if self.options.universe is not None and self.options.universe > 0:
             self.mpiexec_options.append("-np %s" % self.options.universe)
         else:
-            self.mpiexec_options.append("-np %s" % (self.mpiprocesspernode * len(self.nodes)))
+            self.mpiexec_options.append("-np %s" % len(self.nodes))
 
         # pass local env variables to mpiexec
         self.mpiexec_options += self.get_mpiexec_opts_from_env()
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -539,25 +539,26 @@ class MPI(object):
 
         Parses the list of nodes that run an MPI process and writes this information to a nodefile.
         Also parses the list of unique nodes and writes this information to a mpdbootfile
-        (based on hyrda and universe options).
+        (based on hydra and universe options).
         """
         self.make_mympirundir()
 
         if self.mpinodes is None:
             self.set_mpinodes()
 
-        nodetxt = "\n".join(self.mpinodes + [''])
 
         mpdboottxt = ""
-        for node in nub(self.nodes):
+        universe_ppn = self.get_universe_ncpus()
+
+        for node in nub(self.mpinodes):
             txt = node
             if not self.has_hydra:
                 if self.options.universe is not None and self.options.universe > 0:
-                    txt += ":%s" % self.get_universe_ncpus()
+                    txt += ":%s" % universe_ppn[node]
                 txt += " ifhn=%s" % node
-
             mpdboottxt += "%s\n" % txt
 
+        nodetxt = '\n'.join(self.mpinodes)
         nodefn = os.path.join(self.mympirundir, 'nodes')
         mpdfn = os.path.join(self.mympirundir, 'mpdboot')
         try:
@@ -573,8 +574,21 @@ class MPI(object):
             self.log.raiseException(msg)
 
     def get_universe_ncpus(self):
-        """Return ppn for universe"""
-        return self.options.universe
+        """Return ppn dict for universe"""
+        universe_ppn = {}
+        for node in nub(self.nodes):
+            universe_ppn[node] = 0
+        i = 0
+        while i < self.options.universe:
+            for node, ppn in universe_ppn.items():
+                if ppn < self.ppn_dict[node]:
+                    universe_ppn[node] = ppn+1
+                    i += 1
+                if i >= self.options.universe:
+                    break
+
+        return universe_ppn
+
 
     def make_mympirundir(self):
         """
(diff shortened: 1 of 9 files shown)


boegel:
@JensTimmerman where did the `IPy` installation on Jenkins go? 😡

JensTimmerman:
@boegel it should be in tests_require (or install_require) then setup.py test will install it for you, you shouldn't count on jenkins having the correct python packages installed, but declare them correctly in your setup.py

That said, if you look at the logs, this is what happens, so it seems some path is broken somewhere?

stdweird:
@Caylo in end2end you run a subshell with modified PYTHONPATH, you need to include the `topdir/.eggs` also (it's where setuptools will install its dependencies during testing).

boegel:
@stdweird OK fine, but why is this broken all of a sudden? Something must have changed on Jenkins, since #92 and #91 were testing just fine (and the `end2end` tests are not being touched here).

What am I missing?

JensTimmerman:
#92 ran with vsc-install 0.10.9, this now has vsc-install 0.10.19, this is something that changed.

test/end2end.py
```diff
@@ -64,7 +64,9 @@ def setUp(self):
         # add /bin to $PATH, /lib to $PYTHONPATH
         self.topdir = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
         self.mympiscript = os.path.join(os.path.join(self.topdir, 'bin'), 'mympirun.py')
-        os.environ['PYTHONPATH'] = '%s:%s' % (os.path.join(self.topdir, 'lib'), os.getenv('PYTHONPATH', ''))
+        lib = os.path.join(self.topdir, 'lib')
+        eggs = os.path.join(self.topdir, '.eggs')
```
boegel: maybe this should be the `lib/python<version>/site-packages` subdir?

to figure out the version, use `sys.version` (and drop the last digit, i.e. use `2.7`, not `2.7.10`)

stdweird: no it's not

test/end2end.py
```diff
@@ -64,8 +65,10 @@ def setUp(self):
         # add /bin to $PATH, /lib to $PYTHONPATH
         self.topdir = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
         self.mympiscript = os.path.join(os.path.join(self.topdir, 'bin'), 'mympirun.py')
-        os.environ['PYTHONPATH'] = '%s:%s' % (os.path.join(self.topdir, 'lib'), os.getenv('PYTHONPATH', ''))
-
+        lib = os.path.join(self.topdir, 'lib')
+        pyver = '%s.%s' % (sys.version_info.major, sys.version_info.minor)
```
boegel: not used, so please drop

test/end2end.py
```diff
@@ -64,8 +65,10 @@ def setUp(self):
...
+        eggs = ':'.join(glob.glob(os.path.join(self.topdir, '.eggs', '*.egg')))
+        os.environ['PYTHONPATH'] = '%s:%s:%s' % (eggs, lib, os.getenv('PYTHONPATH', ''))
```
boegel: please print the value of `$PYTHONPATH` for debugging (although it looks correct)

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -67,7 +67,7 @@ def pinning_override(self):
 
         ranktxt = ""
         sockets_per_node = 2
-        universe = self.options.universe if self.options.universe else (self.mpiprocesspernode * self.nruniquenodes)
+        universe = self.options.universe if self.options.universe else len(self.nodes)
```
boegel:
```python
universe = self.options.universe or len(self.nodes)
```

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -67,7 +67,7 @@ def pinning_override(self):
 
         ranktxt = ""
         sockets_per_node = 2
```
boegel: hmm, hardcoding much? why 2?

lib/vsc/mympirun/rm/sched.py
```diff
@@ -106,11 +106,15 @@ def __init__(self, options=None, **kwargs):
...
+        self.multiplier = None
+        self.set_multiplier()
 
-        self.mpiprocesspernode = None
-        self.set_mpiprocesspernode()
+        ppn = os.environ.get('PBS_NUM_PPN')
+        if ppn is not None:
+            self.ppn = int(ppn)
+        else:
+            LOGGER.error('PPN is None')
```
boegel: this will just print an error, not raise an exception

also, mention `$PBS_NUM_PPN`: `Failed to determine # cores per node (ppn) via $PBS_NUM_PPN`

lib/vsc/mympirun/rm/sched.py
```diff
@@ -192,10 +196,12 @@ def set_nodes(self):
...
     def set_ppn(self):
         """Determine the processors per node, based on the list of nodes and the list of unique nodes"""
-
-        self.ppn = len(self.nodes) // len(nub(self.nodes))
-
-        self.log.debug("Set ppn to %s", self.ppn)
+        self.ppn_dict = {}
+        for node in self.nodes:
+            if node in self.ppn_dict:
+                self.ppn_dict[node] = self.ppn_dict[node]+1
```
boegel:
```python
for node in self.nodes:
    self.ppn_dict.setdefault(node, 0)
    self.ppn_dict[node] += 1
```

lib/vsc/mympirun/rm/sched.py
```diff
@@ -266,8 +258,11 @@ def set_mpinodes(self):
         if getattr(self.options, 'double', False):
             res = self.nodes * 2
         else:
-            for uniquenode in nub(self.nodes):
-                res.extend([uniquenode] * self.mpiprocesspernode)
+            if getattr(self.options, 'hybrid', None):
```
boegel: isn't `self.options.hybrid` always going to be there?

test/end2end.py
```diff
@@ -64,14 +66,16 @@ def setUp(self):
         # add /bin to $PATH, /lib to $PYTHONPATH
         self.topdir = os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
         self.mympiscript = os.path.join(os.path.join(self.topdir, 'bin'), 'mympirun.py')
-        os.environ['PYTHONPATH'] = '%s:%s' % (os.path.join(self.topdir, 'lib'), os.getenv('PYTHONPATH', ''))
-
+        lib = os.path.join(self.topdir, 'lib')
+        eggs = ':'.join(glob.glob(os.path.join(self.topdir, '.eggs', '*.egg')))
+        os.environ['PYTHONPATH'] = '%s:%s:%s' % (eggs, lib, os.getenv('PYTHONPATH', ''))
```
boegel: add a comment to explain why you're doing this

test/end2end.py
```diff
@@ -64,14 +66,16 @@ def setUp(self):
...
         self.tmpdir = tempfile.mkdtemp()
 
         # make sure we're using the right mympirun installation...
         ec, out = run_simple("%s -c 'import vsc.mympirun; print vsc.mympirun.__file__'" % sys.executable)
         expected_path = os.path.join(self.topdir, 'lib', 'vsc', 'mympirun')
         self.assertTrue(os.path.samefile(os.path.dirname(out.strip()), expected_path))
+        set_env()
```
boegel: also here, add a comment to explain why you're doing this

test/sched.py
```diff
@@ -49,14 +50,25 @@
 
 os.environ['PBS_JOBID'] = "1"
 
-PBSNODEFILE = tempfile.NamedTemporaryFile(delete=False)
-PBSNODEFILE.write("localhost\nlocalhost\n")
-PBSNODEFILE.close()
-os.environ['PBS_NODEFILE'] = PBSNODEFILE.name
+def set_env():
```
boegel: rename as `set_PBS_env`?

test/sched.py
```diff
@@ -49,14 +50,25 @@
...
 class TestSched(unittest.TestCase):
     """tests for vsc.mympirun.mpi.sched functions"""
 
+    def setUp(self):
+        self.orig_environ = os.environ
+        set_env()
+
+    def tearDown(self):
+        """Clean up after running test."""
+        os.environ = self.orig_environ
```
boegel: also clean up the file in `$PBS_NODEFILE`, maybe have a `cleanup_PBS_env` too, which accepts `self.orig_environ` as parameter

test/sched.py
```diff
@@ -105,3 +117,33 @@ def test_get_node_list(self):
...
+        nodefile.write(text)
+        nodefile.close()
+
+        inst = getinstance(mpim.MPI, pbs_class, MympirunOption())
+        inst.set_mpinodes()
+        self.assertEqual(inst.mpinodes, nodes)
+
+        inst.options.double = True
+        inst.set_mpinodes()
+        self.assertEqual(inst.mpinodes, nodes + nodes)
```
boegel: add a clear comment above each block to clarify which aspect your verifying exactly

Caylo:
TODO:
- check if  this works as it should for `double`, `hybrid` and `universe`
- check `machinefile` option for openmpi + tests for openmpi

lib/vsc/mympirun/rm/sched.py
Caylo: yeah, but it defaults to `None` when it isn't set explicitly, so this check will be False

boegel: `if self.options.hybrid is not None`?

lib/vsc/mympirun/mpi/openmpi.py
Caylo: this should be fixed but not in this PR: i made an issue https://github.com/hpcugent/vsc-mympirun/issues/97

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -572,8 +574,21 @@ def make_node_file(self):
             self.log.raiseException(msg)
 
     def get_universe_ncpus(self):
```
boegel: @Caylo please add tests for this method, should be fairly easy to test in isolation?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -572,8 +574,21 @@ def make_node_file(self):
...
-        """Return ppn for universe"""
-        return self.mpiprocesspernode
+        """Return ppn dict for universe"""
```
boegel: @Caylo better: `Construct dictionary with number of processes to start per node, based on --universe`

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -572,8 +574,21 @@ def make_node_file(self):
...
+        for node in nub(self.nodes):
+            universe_ppn[node] = 0
+        i = 0
+        while i < self.options.universe:
+            for node, ppn in universe_ppn.items():
+                if ppn < self.ppn_dict[node]:
+                    universe_ppn[node] = ppn+1
+                    i += 1
+                if i >= self.options.universe:
+                    break
```
boegel: not a big fan of basically having the same condition in two places (here + in `while`)...

maybe something like this?

but write a test first for your current implementation :)

```python
nodes = nub(self.nodes)
universe_ppn = dict((node, 0) for node in nodes)
proc_cnt = 0
node = nodes.pop(0)
while proc_cnt < self.options.universe:
    if universe_ppn[node] < self.ppn_dict[node]:
        universe_ppn[node] += 1
        proc_cnt += 1
        # recycle node
        nodes.append(node)
    # select next node to assign a process to
    node = nodes.pop(0)
```

You'll need to be careful with the last bit, in case `nodes` becomes empty (nodes are only recycled if a process was assigned).

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -778,16 +797,20 @@ def set_mpiexec_options(self):
...
 
         # pass local env variables to mpiexec
         self.mpiexec_options += self.get_mpiexec_opts_from_env()
 
     def make_mpiexec_hydra_options(self):
         """Hydra specific mpiexec options."""
         self.get_hydra_info()
-        self.mpiexec_options.append("--hostfile %s" % self.mpiexec_node_filename)
+        # see https://software.intel.com/en-us/articles/controlling-process-placement-with-the-intel-mpi-library
+        # machinefile keeps the imbalance if there is one; hostfile doesn't
```
boegel: mention `--machinefile` and `--hostfile` to make it clear you're referring to options

lib/vsc/mympirun/rm/sched.py
```diff
@@ -106,11 +106,16 @@ def __init__(self, options=None, **kwargs):
...
-        self.ppn = None
-        self.set_ppn()
+        self.multiplier = None
+        self.set_multiplier()
 
-        self.mpiprocesspernode = None
-        self.set_mpiprocesspernode()
+        ppn = os.environ.get('PBS_NUM_PPN')
+        if ppn is not None:
+            self.ppn = int(ppn)
```
boegel: please also log here

lib/vsc/mympirun/rm/sched.py
```diff
@@ -222,37 +227,23 @@ def is_large(self):
...
         hybrid = getattr(self.options, 'hybrid', None)
         double = getattr(self.options, 'double', False)
 
-        # set the multiplier
         if hybrid:
-            multi = hybrid
+            self.multiplier = hybrid
         elif double:
-            multi = 2
+            self.multiplier = 2
```
boegel: this is kind of weird, you're using two different mechanism here; one for `--hybrid` where `self.multiplier` is the number of processes to stat per 'unit' of cores (#ppn), and another for `--double` where `self.multiplier` is `2`...

Can't you just set `self.multiplier` to `#ppn * 2` in the case of `--double`, and avoid treating it as a special case compared to `--hybrid`? Maybe I'm missing something...

boegel:
@Caylo missing tests for `--universe` + an environment where `$PBS_NUM_PPN` is defined (maybe more)

lib/vsc/mympirun/rm/sched.py
stdweird: this is intended, specifying `--hybrid` and `--double` means `use hybrid as multiplier`

boegel: @stdweird ah, so `--double` means `2* #hybrid`, where the default for `hybrid` is `ppn`?

In that case: OK, but let's clarify that with a comment. :)

Caylo: fix this later, see #98

test/mpi.py
```diff
@@ -300,3 +300,25 @@ def test_fake_dirname(self):
...
+        inst.nodes = [
+            'node1',
+            'node1',
+            'node2',
+            'node2',
+            'node2',
+            'node2',
+        ]
+        options = {
+            3: {'node1': 2, 'node2': 1},
```
boegel: also test `1`? `0`?

boegel:
Thanks @Caylo!

PR #101, opened 2016-12-09
introduce --multi option

### LINKED ISSUE #98: fix how --double and --hybrid work together
right now using `--double` and `--hybrid`will use the hybrid option as a multiplier
what it has to be:
- give `--double` an argument (store_or_None)
- conflict `--double`and `--hybrid`

Caylo:
fixes #98


lib/vsc/mympirun/rm/sched.py
```diff
@@ -256,14 +251,11 @@ def set_mpinodes(self):
         """
 
         res = []
-        if getattr(self.options, 'double', False):
-            res = self.nodes * 2
+        if self.options.hybrid is not None:
```
boegel: @Caylo positive logic is easier to follow, so use `if self.options.hybrid is None` and flip `if`/`else` around

PR #102, opened 2016-12-14
fix bug where vars are passed + test

Caylo:


test/end2end.py
```diff
@@ -261,6 +264,29 @@ def test_option_universe(self):
         # re-set pbs environment
         set_PBS_env()
 
+
+    def test_env_variables(self):
+        """ Test the passing of (extra) variables """
+        fake_mpirun_env = """#!/bin/bash
+        echo 'fake mpirun called with args:' $@
+        echo `env`
```
boegel: just use `env`?

test/end2end.py
```diff
@@ -261,6 +264,29 @@ def test_option_universe(self):
...
+            "hostname",
+        ])
+        ec, out = run_simple(command)
+
+        for key in nub(filter(os.environ.has_key, MPI.OPTS_FROM_ENV_BASE)):
+            self.assertTrue(key in out, "%s is not in out" % key)
+        regex = r'.*-envlist [^ ]USER.*'
+        self.assertTrue(regex.find(out), "Couldn't find variablesprefix USER in script env")
+
+
```
boegel: please also check that `PYTHONPATH` is being passed (if it is defined in the environment), since this is what caused us to find the bug

test/mpi.py
```diff
@@ -260,7 +260,7 @@ def test_set_mpiexec_opts_from_env(self):
 
         print("opts_from_env: %s" % mpi_instance.mpiexec_opts_from_env)
         for env_var in mpi_instance.mpiexec_opts_from_env:
-            self.assertTrue(env_var.startswith(tuple(prefixes)),
+            self.assertTrue(env_var.startswith(tuple(prefixes)) or env_var in mpi_instance.OPTS_FROM_ENV_BASE,
```
boegel: this is only checking in one direction, maybe also add a couple of checks like:

```python
self.assertTrue('PYTHONPATH' in mpi_instance.mpiexec_opts_from_env)
```

You'll need to make sure `PYTHONPATH` is defined in `os.environ` before running `set_mpiexec_opts_from_env()`; also make sure you restore the environment in `tearDown`

test/end2end.py
```diff
@@ -283,8 +283,14 @@ def test_env_variables(self):
 
         for key in nub(filter(os.environ.has_key, MPI.OPTS_FROM_ENV_BASE)):
             self.assertTrue(key in out, "%s is not in out" % key)
-        regex = r'.*-envlist [^ ]USER.*'
-        self.assertTrue(regex.find(out), "Couldn't find variablesprefix USER in script env")
+
+        regex = r'.*-envlist [^ ]*USER.*'
+        self.assertTrue(regex.find(out), "Variablesprefix USER isn't passed to mympirun script env")
+
+        if os.environ.has_key('PYTHONPATH'):
```
boegel: no, hard set `PYTHONPATH`, so you know what you're checking:

```python
os.environ['PYTHONPATH'] = '/just/an/example:%s' % os.getenv('PYTHONPATH', '')
...
```

test/mpi.py
```diff
@@ -253,6 +260,10 @@ def test_set_mpiexec_global_options(self):
     def test_set_mpiexec_opts_from_env(self):
         """test if mpiexec_opts_from_env only contains environment variables that start with the given prefix"""
         mpi_instance = getinstance(mpim.MPI, Local, MympirunOption())
+        check_keys = ['PATH', 'PYTHONPATH']
+        for key in check_keys:
+            if not os.environ.has_key(key):
+                os.environ[key] = "fake %s" % key
```
boegel: at least make the value look like a `(PYTHON)PATH` value... `/example:/test/123`

PR #104, opened 2017-02-15
fix mpiexec options for hybrid option

Caylo:
we probably need to test this better in mympirun itself

### DIFF
--- a/test/mpi.py
+++ b/test/mpi.py
@@ -340,6 +340,28 @@ class TestMPI(TestCase):
             universe_ppn = inst.get_universe_ncpus()
             self.assertEqual(universe_ppn, options[opt])
 
+    def test_get_hybrid_ncpus(self):
+        inst = getinstance(mpim.MPI, Local, MympirunOption())
+        inst.nodes = [
+            'node1',
+            'node1',
+            'node1',
+            'node1',
+            'node2',
+            'node2',
+            'node2',
+            'node2',
+        ]
+        options = range(1,5)
+        for opt in options:
+            inst.options.hybrid = opt
+            inst.set_ppn()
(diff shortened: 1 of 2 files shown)


boegel:
@Caylo +1 on catching this in the mympirun unit tests...

test/mpi.py
```diff
@@ -340,6 +340,28 @@ def test_get_universe_ncpus(self):
             universe_ppn = inst.get_universe_ncpus()
             self.assertEqual(universe_ppn, options[opt])
 
+    def test_get_hybrid_ncpus(self):
+        inst = getinstance(mpim.MPI, Local, MympirunOption())
```
boegel: @Caylo please include a docstring to clarify what you're testing exactly

test/mpi.py
```diff
@@ -340,6 +340,28 @@ def test_get_universe_ncpus(self):
...
+        inst.nodes = [
```
boegel: maybe collapse to this (which is more clear as well):

```python
inst.nodes = ['node1'] * 4 + ['node2'] * 4 
```

test/mpi.py
```diff
@@ -340,6 +340,28 @@ def test_get_universe_ncpus(self):
...
+            'node1',
+            'node1',
+            'node1',
+            'node1',
+            'node2',
+            'node2',
+            'node2',
+            'node2',
+        ]
+        options = range(1,5)
```
boegel: why this range specifically?

Caylo: from 1 to the ppn in the fake nodefile

boegel:
lgtm, thanks @Caylo!

PR #107, opened 2017-03-01
use local scheduler when PBS_NODEFILE is not available

### LINKED ISSUE #106: use local scheduler when PBS_NODEFILE is not available
right now it causes an error
```
$ mympirun hostname
2017-02-22 17:07:29,771 WARNING    mympirun.IntelHydraMPI_PBS MainThread  set_nodes: failed to get PBS_NODEFILE from environment
2017-02-22 17:07:29,772 ERROR      mympirun        MainThread  Main failed
Traceback (most recent call last):
  File "/vscmnt/gent_vulpix/_/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/EGG-INFO/scripts/mympirun", line 123, in main
    instance = getinstance(*instance_options)
  File "/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/vsc/mympirun/factory.py", line 66, in getinstance
    return coupler_class(options=options.options, cmdargs=options.args)
  File "/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/vsc/mympirun/factory.py", line 58, in __init__
    super(Coupler, self).__init__(**kwargs)
  File "/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/vsc/mympirun/mpi/mpi.py", line 310, in __init__
    super(MPI, self).__init__(**kwargs)
  File "/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/vsc/mympirun/rm/sched.py", line 107, in __init__
    self.set_nodes()
  File "/vscmnt/gent_vulpix/_/user/data/gent/vsc413/vsc41353/mympirun-dev/software/vsc-mympirun/dev/lib/python2.7/site-packages/vsc_mympirun-3.4.4-py2.7.egg/vsc/mympirun/rm/pbs.py", line 49, in set_nodes
    self.log.raiseException("set_nodes: failed to get %s from environment"  % nodevar)
  File "/usr/lib/python2.7/site-packages/vsc/utils/fancylogger.py", line 329, in raiseException
    raise exception, message, tb
Exception: set_nodes: failed to get PBS_NODEFILE from environment
```

Caylo:
fixes #106

### DIFF
--- a/lib/vsc/mympirun/rm/sched.py
+++ b/lib/vsc/mympirun/rm/sched.py
@@ -52,11 +52,6 @@ def what_sched(requested):
                 return sched, found_sched
         LOGGER.warn("%s scheduler was requested, but mympirun failed to find an implementation", requested)
 
-    # if PBS_NODEFILE is not present, use local scheduler
-    if not os.getenv('PBS_NODEFILE'):
-        LOGGER.debug("PBS_NODEFILE not found, using local scheduler")
-        return local_sched, found_sched
-
     # next, try to use the scheduler defined by environment variables
     for sched in found_sched:
         if sched.SCHED_ENVIRON_ID in os.environ:
@@ -69,9 +64,12 @@ def what_sched(requested):
 
 def get_local_sched(found_sched):
     """Helper function to get local scheduler (or None, if there is no local scheduler)"""
+    res = None
     for sched in found_sched:
         if sched._is_sched_for("local"):
-            return sched
+            res = sched
+            break
+    return res
 
 
 class Sched(object):
--- a/lib/vsc/mympirun/rm/sched.py
+++ b/lib/vsc/mympirun/rm/sched.py
@@ -42,6 +42,9 @@ def what_sched(requested):
     # The coupler is also a subclass of sched, but we don't want it
     found_sched = [x for x in get_subclasses(Sched) if x.__name__ != 'Coupler']
 
+    # Get local scheduler
+    local_sched = get_local_sched(found_sched)
+
     # first, try to use the scheduler that was requested
     if requested:
         for sched in found_sched:
@@ -50,12 +53,9 @@ def what_sched(requested):
         LOGGER.warn("%s scheduler was requested, but mympirun failed to find an implementation", requested)
 
     # if PBS_NODEFILE is not present, use local scheduler
-    filename = os.environ.get('PBS_NODEFILE', None)
-    if filename is None:
-        for sched in found_sched:
-            if sched._is_sched_for("local"):
-                LOGGER.debug("PBS_NODEFILE not found, using local scheduler")
-                return sched, found_sched
+    if not os.getenv('PBS_NODEFILE'):
+        LOGGER.debug("PBS_NODEFILE not found, using local scheduler")
+        return local_sched, found_sched
 
     # next, try to use the scheduler defined by environment variables
     for sched in found_sched:
@@ -63,13 +63,15 @@ def what_sched(requested):
             return sched, found_sched
 
     # If that fails, try to force the local scheduler
-    for sched in found_sched:
-        LOGGER.debug("No scheduler found in environment, trying local")
-        if sched._is_sched_for("local"):
-            return sched, found_sched
+    LOGGER.debug("No scheduler found in environment, trying local")
+    return local_sched, found_sched
 
-    # if there is no local scheduler, return None
-    return None, found_sched
+
+def get_local_sched(found_sched):
+    """Helper function to get local scheduler (or None, if there is no local scheduler)"""
+    for sched in found_sched:
+            if sched._is_sched_for("local"):
+                return sched
 
 
 class Sched(object):
--- a/lib/vsc/mympirun/rm/sched.py
+++ b/lib/vsc/mympirun/rm/sched.py
@@ -70,8 +70,8 @@ def what_sched(requested):
 def get_local_sched(found_sched):
     """Helper function to get local scheduler (or None, if there is no local scheduler)"""
     for sched in found_sched:
-            if sched._is_sched_for("local"):
-                return sched
+        if sched._is_sched_for("local"):
+            return sched
 
 
 class Sched(object):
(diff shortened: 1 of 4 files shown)


lib/vsc/mympirun/rm/sched.py
```diff
@@ -49,6 +49,14 @@ def what_sched(requested):
                 return sched, found_sched
         LOGGER.warn("%s scheduler was requested, but mympirun failed to find an implementation", requested)
 
+    # if PBS_NODEFILE is not present, use local scheduler
+    filename = os.environ.get('PBS_NODEFILE', None)
+    if filename is None:
```
boegel: only used in one place, so no need for a variable

also, `os.environ.get` uses `None` as fallback value by default, no need to specify it

and usually `os.getenv` is used instead (same behaviour)

```python
if not os.getenv('PBS_NODEFILE'):
```

lib/vsc/mympirun/rm/sched.py
```diff
@@ -49,6 +49,14 @@ def what_sched(requested):
...
+        for sched in found_sched:
+            if sched._is_sched_for("local"):
+                LOGGER.debug("PBS_NODEFILE not found, using local scheduler")
+                return sched, found_sched
```
boegel: this same loop is also used below, please don't duplicate code but factor it out in a function you call out to

lib/vsc/mympirun/rm/sched.py
boegel: also, we should only be doing this if no scheduler was specified (or if the default `pbs` scheduler was used)?

test/end2end.py
```diff
@@ -308,3 +309,14 @@ def change_env(self, cores):
         pbsnodefile.write('\n'.join(['localhost'] * cores))
         pbsnodefile.close()
         os.environ['PBS_NODEFILE'] = pbsnodefile.name
+
+
+    def test_unset_nodefile(self):
+        self.assertTrue(schedm.what_sched(False)[0].__name__ == 'PBS')
```
boegel: test with `isinstance(schedm.what_sched(False), PBS)` ?

test/end2end.py
```diff
@@ -308,3 +309,14 @@ def change_env(self, cores):
...
+        nodefile = os.environ['PBS_NODEFILE']
+        del os.environ['PBS_NODEFILE']
+        # fall back to local if PBS_NODEFILE is not available
+        self.assertTrue(schedm.what_sched(False)[0].__name__ == 'Local')
```
boegel: test with `isinstance`?

test/end2end.py
```diff
@@ -308,3 +309,14 @@ def change_env(self, cores):
...
+
+        # reset nodefile for proper test cleanup
+        os.environ['PBS_NODEFILE'] = nodefile
```
boegel: I don't think this is needed, the tests already restore the environment via `tearDown`? do double check...

lib/vsc/mympirun/rm/sched.py
```diff
@@ -42,27 +42,36 @@ def what_sched(requested):
...
-        LOGGER.debug("No scheduler found in environment, trying local")
-        if sched._is_sched_for("local"):
-            return sched, found_sched
+    LOGGER.debug("No scheduler found in environment, trying local")
+    return local_sched, found_sched
 
 
-    # if there is no local scheduler, return None
-    return None, found_sched
+def get_local_sched(found_sched):
```
boegel: new function, new dedicated test?

try to also test the return of `None` (which may not be easy)

Also, I'd prefer this:

```python
res = None
for sched in found_sched:
    if sched._is_sched_for('local'):
        res = sched
        break
return res

File: lib/vsc/mympirun/rm/sched.py [hunk 7]
```diff
@@ -42,27 +42,36 @@ def what_sched(requested):
...
     if requested:
         for sched in found_sched:
             if sched._is_sched_for(requested):
                 return sched, found_sched
         LOGGER.warn("%s scheduler was requested, but mympirun failed to find an implementation", requested)
 
+    # if PBS_NODEFILE is not present, use local scheduler
+    if not os.getenv('PBS_NODEFILE'):
+        LOGGER.debug("PBS_NODEFILE not found, using local scheduler")
+        return local_sched, found_sched
```
boegel: hmm, looking at the logic that is there now, why is this actually needed?

Isn't the loop below supposed to only return a `PBS` scheduler in case we're actually in a PBS job?

It does check for `$PBS_JOBID` rather than `$PBS_NODEFILE` though, but it seems to suggest the actual bug is somewhere else, and we're cheating here?

test/end2end.py
Caylo: the cleanup_PBS_env in tearDown tries to `os.remove(os.environ['PBS_NODEFILE'])`, which fails if PBS_NODEFILE isn't present

boegel: rather than using `cleanup_PBS_env`, we should just restore the original environment in `tearDown` using

```python
os.environ = self.orig_environ
```

so, let's trash `cleanup_PBS_env`?

test/end2end.py
```diff
@@ -308,3 +311,14 @@ def change_env(self, cores):
...
+
+    def test_unset_nodefile(self):
+        self.assertEqual(schedm.what_sched(False)[0], PBS)
+        nodefile = os.environ['PBS_NODEFILE']
+        del os.environ['PBS_NODEFILE']
+        # fall back to local if PBS_NODEFILE is not available
+        self.assertEqual(schedm.what_sched(False)[0], Local)
+
+        # reset nodefile for proper test cleanup
+        os.environ['PBS_NODEFILE'] = nodefile
```
boegel: this should not be needed, get rid of `cleanup_PBS_env` instead in tearDown and just restore `self.orig_environ`

boegel:
lgtm, thanks @Caylo!

PR #109, opened 2017-03-01
fix nodefile path and environment variables

Caylo:
test are still failing, will be continued next week

edit (by @boegel): proper fix for #106, since ~~#107~~ introduced a bug (`SCHED_ENVIRON_ID` is used to determine MPI base path, and is not expected to have a path value)


lib/vsc/mympirun/rm/pbs.py
```diff
@@ -34,7 +34,8 @@ class PBS(Sched):
 
     """Torque/PBS based"""
     _sched_for = ['pbs', 'torque']
-    SCHED_ENVIRON_ID = 'PBS_NODEFILE'
+    SCHED_ENVIRON_ID = 'PBS_JOBID'
+    SCHED_NODEFILE = 'PBS_NODEFILE'
```
boegel: maybe rename to `SCHED_ENVIRON_NODEFILE`?

test/sched.py
```diff
@@ -53,13 +55,20 @@
...
 def set_PBS_env():
     """ Set up the environment to recreate being in a hpc job """
-    pbsnodefile = tempfile.NamedTemporaryFile(delete=False)
+    if not os.path.isdir('/tmp/%s' % os.environ['USER']):
+        os.mkdir('/tmp/%s' % os.environ['USER'])
+    pbsnodefile = tempfile.NamedTemporaryFile(dir='/tmp/%s' % os.environ['USER'], delete=False)
     pbsnodefile.write("localhost\nlocalhost\n")
     pbsnodefile.close()
     os.environ['PBS_NODEFILE'] = pbsnodefile.name
+    os.chmod(pbsnodefile.name, stat.S_IRUSR)
```
boegel: include a comment to explain why we do this (i.e., in a real situation the file and location of `$PBS_NODEFILE` are not writeable either)

test/sched.py
```diff
@@ -53,13 +55,20 @@
...
+    os.chmod(os.path.dirname(pbsnodefile.name), stat.S_IRUSR|stat.S_IXUSR)
+
 
 def cleanup_PBS_env(orig_env):
     """ cleanup the mock job environment """
+    os.chmod(pbsnodefile.name, stat.S_IWUSR)
+    os.chmod(os.path.dirname(pbsnodefile.name), stat.S_IWUSR|stat.S_IXUSR)
     os.remove(os.environ['PBS_NODEFILE'])
```
boegel: aah, I overlooked that this actually removes to temporary file, together with resetting the environment >_<

so, we should revert the changes made in #107 where we drop the use of `cleanup_PBS_env` to ensure that the temporary file is cleaned up properly

test/sched.py
```diff
@@ -53,13 +55,20 @@
 
 def set_PBS_env():
     """ Set up the environment to recreate being in a hpc job """
-    pbsnodefile = tempfile.NamedTemporaryFile(delete=False)
+    if not os.path.isdir('/tmp/%s' % os.environ['USER']):
+        os.mkdir('/tmp/%s' % os.environ['USER'])
+    pbsnodefile = tempfile.NamedTemporaryFile(dir='/tmp/%s' % os.environ['USER'], delete=False)
```
boegel: better:

```python
tmpdir = tempfile.mkdtemp()
pbsnodefile = os.path.join(pbsnodefile)
...
# make $PBS_NODEFILE and directory it is in read-only, just like in the real world
os.chmod(pbsnodefile, stat.S_IRUSR)
# make location directory where $PBS_NODEFILE resides read-only
os.chmod(tmpdir, stat.S_IRUSR|stat.S_IXUSR)

File: test/sched.py (hunk 3 above)
Comment (boegel) [inline test/sched.py]: change to:

```python
shutil.rmtree(os.path.dirname(os.environ['PBS_NODEFILE']))
```

test/sched.py
```diff
@@ -136,6 +145,7 @@ def test_set_node_list(self):
         ]
         pbs_class = SCHEDDICT['pbs']
         text = '\n'.join(nodes)
+        os.chmod(pbsnodefile.name, stat.S_IRUSR | stat.S_IWUSR)
```
boegel: make it read-only again after we're done re-writing it

test/sched.py
```diff
@@ -53,13 +55,20 @@
...
     pbsnodefile.write("localhost\nlocalhost\n")
     pbsnodefile.close()
     os.environ['PBS_NODEFILE'] = pbsnodefile.name
+    os.chmod(pbsnodefile.name, stat.S_IRUSR)
+    os.chmod(os.path.dirname(pbsnodefile.name), stat.S_IRUSR|stat.S_IXUSR)
+
 
 def cleanup_PBS_env(orig_env):
     """ cleanup the mock job environment """
+    os.chmod(pbsnodefile.name, stat.S_IWUSR)
```
boegel: `pbsnodefile` is not defined in this scope, use `os.environ['PBS_NODEFILE']` (and explain why you're chmod'ing it)

PR #110, opened 2017-03-01
use pbsdsh

Caylo:


lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +831,24 @@ def make_mpiexec_hydra_options(self):
...
-                    self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
-
-            launcher_exec = self.HYDRA_LAUNCHER_EXEC
-            if not launcher:
-                launcher_exec = self.get_rsh()
+        launcher = None
+        if getattr(self, 'HYDRA_LAUNCHER', None) is not None:
```
boegel: positive logic is easier:

```python
launcher = getattr(self, 'HYDRA_LAUNCHER', None)
if launcher:

File: lib/vsc/mympirun/mpi/mpi.py [hunk 2]
```diff
@@ -831,25 +831,24 @@ def make_mpiexec_hydra_options(self):
...
+            launcher = [x for x in self.HYDRA_LAUNCHER if x in self.hydra_info.get('launcher', [])]
+            if launcher:
+                self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
             else:
-                self.mpiexec_options.append("-%s %s" % (self.HYDRA_LAUNCHER_NAME, launcher[0]))
+                self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
+                                self.HYDRA_LAUNCHER, self.hydra_info)
+
+        launcher_exec = self.HYDRA_LAUNCHER_EXEC
```
boegel: also use `getattr` with a fallback value here?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +831,24 @@ def make_mpiexec_hydra_options(self):
...
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
-
-            launcher_exec = self.HYDRA_LAUNCHER_EXEC
-            if not launcher:
-                launcher_exec = self.get_rsh()
+        launcher = None
+        if getattr(self, 'HYDRA_LAUNCHER', None) is not None:
+            launcher = [x for x in self.HYDRA_LAUNCHER if x in self.hydra_info.get('launcher', [])]
```
boegel: use `launcher` rather than `self.HYDRA_LAUNCHER`?

also, redefining `launcher` here makes this harder to understand...

boegel:
@Caylo still missing: support for `--launcher` cmdline option (I know this is still WIP, kind of a reminder to self ;))

Caylo:
todo: 
- warn when --launcher option doesnt exist
- end2end tests

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
             else:
                 self.log.debug("make_mpiexec_hydra_options: no rmk from HYDRA_RMK %s and hydra_info %s",
                                self.HYDRA_RMK, self.hydra_info)
+
+        default_launcher = getattr(self, 'HYDRA_LAUNCHER', None)
+        if self.options.launcher:
+            launcher = [self.options.launcher]
+            launcher.extend(default_launcher)
```
boegel: I wouldn't use the default launcher as a fallback; if `--launcher` is specified, use that, and nothing else

don't try to outwit the user making a typo here, there *may* be a good reason to specify an unknown launcher

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
...
-            if getattr(self, 'HYDRA_LAUNCHER', None) is not None:
-                launcher = [x for x in self.HYDRA_LAUNCHER if x in self.hydra_info.get('launcher', [])]
-                if launcher:
-                    self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
+            launcher = default_launcher
+
+        if launcher:
```
boegel: at least print a debug statement in an `else` here that says something like `No custom launcher specified`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
...
         else:
-            launcher = None
-            if getattr(self, 'HYDRA_LAUNCHER', None) is not None:
-                launcher = [x for x in self.HYDRA_LAUNCHER if x in self.hydra_info.get('launcher', [])]
-                if launcher:
-                    self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
+            launcher = default_launcher
```
boegel: log debug `Using default launcher ...`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
...
-                self.mpiexec_options.append("-%s %s" % (self.HYDRA_LAUNCHER_NAME, launcher[0]))
+                self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
+                                self.HYDRA_LAUNCHER, self.hydra_info)
+
+                launcher_exec = getattr(self, 'HYDRA_LAUNCHER_EXEC', None)
+
+                if launcher_exec is not None:
+                    self.log.debug("make_mpiexec_hydra_options: HYDRA using launcher exec %s", launcher_exec)
+                else:
+                    launcher_exec = self.get_rsh()
```
boegel: should this block be under the `if launcher`? doesn't it make sense to specify a custom `launcher_exec` without specifying a custom launcher?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
...
-                    self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
+            launcher = default_launcher
+
+        if launcher:
+            # filter out the launchers that are not in self.hydra_info
+            avail_launcher = [x for x in launcher if x in self.hydra_info.get('launcher', [])]
+            if avail_launcher:
```
boegel: I would skip this filtering magic, and just straight up either use the value passed to `--launcher` or whatever `HYDRA_LAUNCHER` specifies?

Or, at the very least, don't just silently skip an unknown specified launcher, be very verbose about it (print a warning). And if so, provide an option to force trying to use an unknown launcher.

Long story short, I wouldn't try to be too smart about this, just use what was specified.

Maybe do check what kind of error is printed by Intel MPI when you try to use an unknown launcher; if it's a very obscure error, maybe that's the reason this was added originally.

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,26 +830,38 @@ def make_mpiexec_hydra_options(self):
...
+                    launcher_exec = self.get_rsh()
 
-            if launcher_exec is not None:
-                self.log.debug("make_mpiexec_hydra_options: HYDRA using launcher exec %s", launcher_exec)
                 self.mpiexec_options.append("-%s-exec %s" % (self.HYDRA_LAUNCHER_NAME, launcher_exec))
 
+
+
+
+
```
boegel: needs more whitespace? ;)

lib/vsc/mympirun/option.py
```diff
@@ -76,6 +76,8 @@ def make_init(self):
 
             "hybrid": ("Run in hybrid mode, specify number of processes per node.", "int", "store", None, 'h'),
 
+            "launcher": ("The launcher to be used by Hydra", "str", "store", None),
```
boegel: maybe mention `ssh`, `pbs` as examples?

lib/vsc/mympirun/rm/pbs.py
```diff
@@ -39,7 +39,7 @@ class PBS(Sched):
 
     RSH_LARGE_CMD = 'pbsssh'
     RSH_LARGE_LIMIT = 'pbsssh'
-    HYDRA_LAUNCHER_EXEC = 'pbsssh'
+    HYDRA_LAUNCHER_EXEC = 'pbsssh' # only used when default pbsdsh is not available (under impi 5.0)
```
boegel: the comment is specified to Intel MPI, so clarify:

```python
# pbsssh is only used if native support for pbsdsh is not supported in the MPI library;
# e.g. Intel MPI v5.x and newer supports using pbsdsh as launcher natively, no need for pbsssh wrapper

File: lib/vsc/mympirun/mpi/mpi.py (hunk 7 above)
Comment (boegel) [inline lib/vsc/mympirun/mpi/mpi.py]: ok, this is probably because of `pbsssh` being the default `launcher_exec`?

you should change that: only use `pbsssh` as default `launcher_exec` if `ssh` is used as a launcher?

File: lib/vsc/mympirun/rm/sched.py [hunk 12]
```diff
@@ -93,7 +93,7 @@ class Sched(object):
     RSH_LARGE_LIMIT = 16
 
     HYDRA_RMK = []
-    HYDRA_LAUNCHER = ['ssh']
+    HYDRA_LAUNCHER = ['pbsdsh']
```
boegel: I would include a version check, and use `pbsdsh` by default for `Intel MPI` v5.x and up, and stick to `ssh` else (e.g. also for other MPI libraries like Open MPI, for now)

That probably means overriding `HYDRA_LAUNCHER` in `intelmpi.py` to `pbsdsh` if v5.x or newer, and leaving the default to be `ssh` like it is now?
That, and make sure you're not using `pbsssh`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +830,39 @@ def make_mpiexec_hydra_options(self):
...
+        launcher = None
+        default_launcher = getattr(self, 'HYDRA_LAUNCHER', None)
+        avail_launchers = self.hydra_info.get('launcher', [])
+
+        if self.options.launcher:
+            launcher = self.options.launcher
+            if launcher not in avail_launchers:
+                err = "Specified launcher %s does not exist, available launchers: %s"
+                self.log.raiseException(err % (launcher, avail_launchers))
```
boegel: Do we want to enforce only using known launcher, or just warn about using what looks like an unknown launcher and then continue anyway? I would do the latter...

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +830,39 @@ def make_mpiexec_hydra_options(self):
...
-                    self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", launcher)
-                else:
-                    self.log.debug("make_mpiexec_hydra_options: no launcher from HYDRA_LAUNCHER %s and hydra_info %s",
-                                   self.HYDRA_LAUNCHER, self.hydra_info)
+            if default_launcher in avail_launchers:
+                self.log.debug("No launcher specified, using default launcher: %s" % default_launcher)
+                launcher = default_launcher
+            elif avail_launchers:
+                launcher = avail_launchers[0]
+                self.log.debug("make_mpiexec_hydra_options: HYDRA: launcher %s, using first one", avail_launchers)
```
boegel: I'm not sure if this is a good idea, even though the original code does this...

Here I would bump out with an error like `"No launcher specified, and no default launcher defined"`

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +830,39 @@ def make_mpiexec_hydra_options(self):
...
-            launcher_exec = self.HYDRA_LAUNCHER_EXEC
-            if not launcher:
-                launcher_exec = self.get_rsh()
-            else:
-                self.mpiexec_options.append("-%s %s" % (self.HYDRA_LAUNCHER_NAME, launcher[0]))
+        if launcher:
+            self.mpiexec_options.append("-%s %s" % (self.HYDRA_LAUNCHER_NAME, launcher))
+
+        # when using ssh launcher, use custom pbsssh wrapper as exec
+        if launcher=='ssh':
```
boegel: spaces around `==`

lib/vsc/mympirun/option.py
```diff
@@ -76,6 +76,8 @@ def make_init(self):
 
             "hybrid": ("Run in hybrid mode, specify number of processes per node.", "int", "store", None, 'h'),
 
+            "launcher": ("The launcher to be used by Hydra (ssh, pbsdsh)", "str", "store", None),
```
boegel: `(for example: ssh, pbsdsh, ...)`

lib/vsc/mympirun/rm/sched.py
```diff
@@ -93,7 +93,7 @@ class Sched(object):
     RSH_LARGE_LIMIT = 16
 
     HYDRA_RMK = []
-    HYDRA_LAUNCHER = ['ssh']
+    HYDRA_LAUNCHER = 'pbsdsh'
```
boegel: this should only be done for (sufficiently recent version of) Intel MPI?

test/end2end.py
```diff
@@ -270,6 +270,7 @@ def test_option_universe(self):
         # intel mpi with hydra
         del os.environ['I_MPI_PROCESS_MANAGER']
         ec, out = run_simple("%s %s --setmpi ihmpirun --universe 1 hostname" % (sys.executable, self.mympiscript))
+        print "--------------- %s" % out
```
boegel: leftover from debugging?

lib/vsc/mympirun/rm/pbs.py
```diff
@@ -37,6 +37,8 @@ class PBS(Sched):
     SCHED_ENVIRON_ID = 'PBS_JOBID'
     SCHED_ENVIRON_NODEFILE = 'PBS_NODEFILE'
 
+    # pbsssh is only used if native support for pbsdsh is not supported in the MPI library;
+    # e.g. Intel MPI v5.x and newer supports using pbsdsh as launcher natively, no need for pbsssh wrapper
```
boegel: for OpenMPI we shouldn't use `pbsssh` either, OpenMPI will use a correct launcher mechanism by default if it was built with Torque integration, see https://www.open-mpi.org/faq/?category=tm

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +830,38 @@ def make_mpiexec_hydra_options(self):
...
+
+        # when using ssh launcher, use custom pbsssh wrapper as exec
+        if launcher == 'ssh':
+            launcher_exec = getattr(self, 'HYDRA_LAUNCHER_EXEC', None)
 
             if launcher_exec is not None:
                 self.log.debug("make_mpiexec_hydra_options: HYDRA using launcher exec %s", launcher_exec)
-                self.mpiexec_options.append("-%s-exec %s" % (self.HYDRA_LAUNCHER_NAME, launcher_exec))
+            else:
+                launcher_exec = self.get_rsh()
```
boegel: what if `get_rsh` returns `None`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,25 +830,38 @@ def make_mpiexec_hydra_options(self):
...
+        avail_launchers = self.hydra_info.get('launcher', [])
+
+        if self.options.launcher:
+            launcher = self.options.launcher
+            if launcher not in avail_launchers:
+                err = "Specified launcher %s does not exist, available launchers: %s"
+                self.log.warning(err % (launcher, avail_launchers))
+
+        else:
+            if default_launcher in avail_launchers:
```
boegel: this should just be `if default_launcher:`?

lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -246,3 +246,11 @@ def set_mpiexec_global_options(self):
             # force it
             self.mpiexec_global_options['I_MPI_FABRICS'] = 'shm:ofa'
             self.mpiexec_global_options['I_MPI_OFA_USE_XRC'] = 1
+
+
+class IntelHydraMPI_ver5(IntelHydraMPI):
```
boegel: rename to `IntelHydraMpiPbsdsh`?

lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -246,3 +246,11 @@ def set_mpiexec_global_options(self):
...
+
+    """ MPI class for IntelMPI, with hydra and supporting pbsdsh """
+    # pbsdsh is supported from 5.0: http://scc.ustc.edu.cn/zlsc/tc4600/intel/2016.0.109/mpi/Release_Notes.txt
+    _mpirun_version = staticmethod(lambda x: LooseVersion(x) >= LooseVersion("5.0.3.0"))
```
boegel: drop the last `.0`

lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -246,3 +246,11 @@ def set_mpiexec_global_options(self):
             # force it
             self.mpiexec_global_options['I_MPI_FABRICS'] = 'shm:ofa'
             self.mpiexec_global_options['I_MPI_OFA_USE_XRC'] = 1
+
+
+class IntelHydraMPI_ver5(IntelHydraMPI):
+
+    """ MPI class for IntelMPI, with hydra and supporting pbsdsh """
+    # pbsdsh is supported from 5.0: http://scc.ustc.edu.cn/zlsc/tc4600/intel/2016.0.109/mpi/Release_Notes.txt
```
boegel: s/5.0/Intel MPI 5.0.3/g

use https://software.intel.com/sites/default/files/managed/b7/99/intelmpi-5.0-update3-releasenotes-linux.pdf as reference

lib/vsc/mympirun/rm/pbs.py
boegel: follow-up in https://github.com/hpcugent/vsc-mympirun/issues/112, but #113 suggest we're already doing the right thing (we should just add a test for it)

test/end2end.py
```diff
@@ -269,7 +277,9 @@ def test_option_universe(self):
 
         # intel mpi with hydra
         del os.environ['I_MPI_PROCESS_MANAGER']
-        ec, out = run_simple("%s %s --setmpi ihmpirun --universe 1 hostname" % (sys.executable, self.mympiscript))
+        # specific launcher specified because there is no hydra_info in this test environment
```
boegel: is this still required now that the fake `mpirun` supports `-info`?

test/end2end.py
```diff
@@ -322,3 +332,23 @@ def test_unset_nodefile(self):
...
+
+        # default behavior
+        ec, out = run_simple("%s %s --setmpi ihmpirun hostname" % (sys.executable, self.mympiscript))
+        regex = r'-bootstrap pbsdsh'
+        self.assertTrue(regex.find(out), "-bootstrap option is not pbsdsh (default)")
+
+        # forced behavior
+        ec, out = run_simple("%s %s --setmpi ihmpirun --launcher ssh hostname" % (sys.executable, self.mympiscript))
+        regex = r'-bootstrap ssh'
+        self.assertTrue(regex.find(out), "-bootstrap option is not ssh (with option)")
```
boegel: also check for `pbsssh` as launcher exec?

lib/vsc/mympirun/option.py
```diff
@@ -76,6 +76,8 @@ def make_init(self):
 
             "hybrid": ("Run in hybrid mode, specify number of processes per node.", "int", "store", None, 'h'),
 
+            "launcher": ("The launcher to be used by Hydra (for example: ssh, pbsdsh, ..)", "str", "store", None),
```
boegel: Do we need to mention that this is only used for recent Intel MPI versions?

lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -246,3 +246,12 @@ def set_mpiexec_global_options(self):
             # force it
             self.mpiexec_global_options['I_MPI_FABRICS'] = 'shm:ofa'
             self.mpiexec_global_options['I_MPI_OFA_USE_XRC'] = 1
+
+
+class IntelHydraMPIPbsdsh(IntelHydraMPI):
+
```
boegel: drop this empty line

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -831,30 +830,42 @@ def make_mpiexec_hydra_options(self):
...
+        launcher = None
+        default_launcher = getattr(self, 'HYDRA_LAUNCHER', None)
+        avail_launchers = self.hydra_info.get('launcher', [])
+
+        if self.options.launcher:
+            launcher = self.options.launcher
+            if launcher not in avail_launchers:
+                err = "Specified launcher %s does not exist, available launchers: %s"
+                self.log.warning(err % (launcher, avail_launchers))
+
```
boegel: same here, too much whitespace, remove this empty line

test/end2end.py
```diff
@@ -322,3 +331,27 @@ def test_unset_nodefile(self):
...
+
+
+    def test_launcher_opt(self):
+        """Test --launcher command line option"""
+        install_fake_mpirun('mpirun', self.tmpdir)
+
+        # default behavior
+        ec, out = run_simple("%s %s --setmpi ihmpirun hostname" % (sys.executable, self.mympiscript))
+        regex = r'-bootstrap pbsdsh'
+        self.assertTrue(regex.find(out), "-bootstrap option is not pbsdsh (default)")
```
boegel: include `out` in the error message so we can see what is there in case the check fails

test/end2end.py
```diff
@@ -322,3 +331,27 @@ def test_unset_nodefile(self):
...
+
+        # forced behavior
+        ec, out = run_simple("%s %s --setmpi ihmpirun --launcher ssh hostname" % (sys.executable, self.mympiscript))
+        regexes = [
+            (r'-bootstrap ssh', "bootstrap option is not ssh (with option)"),
+            (r'-bootstrap-exec pbsssh', "bootstrap-exec is should be pbsssh when specified launcher is ssh")
+        ]
+        for regex in regexes:
+            self.assertTrue(regex[0].find(out), regex[1])
```
boegel: also here, include `out`:

```python
regex[1] + ": " + out
```

test/end2end.py
```diff
@@ -322,3 +331,27 @@ def test_unset_nodefile(self):
...
+
+        # wrong behavior
```
boegel: maybe change to

```
# unknown launcher being specified only results in a warning (we allow specifying launchers that are not listed)
```

lib/vsc/mympirun/mpi/mpi.py
boegel: nevermind for now, the original code doesn't take this into account either, and this is a fallback that will likely never be used

boegel:
@Caylo using `mympirun --sched=local` now results in `-bootstrap ['local']` being passed...

boegel:
lgtm, thanks @Caylo!

PR #120, opened 2017-04-12
fix nodefile when using --universe

### LINKED ISSUE #119: [mympirun v4.0.0 beta] --universe isn't balanced over available nodes
both with the latest version(s) of impi and ompi, the load when using `--universe` is all put on one node instead of dividing the processes over both nodes

Caylo:
fixes #119

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -584,19 +584,25 @@ class MPI(object):
         if self.mpinodes is None:
             self.set_mpinodes()
 
-
-        mpdboottxt = ""
+        nodetxt = ""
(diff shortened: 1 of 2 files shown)


boegel:
@Caylo a test is missing that verifies the behaviour of `--universe` and how processes are spread across nodes?

Also, we'll need to figure out whether we really want to spread processes uniformly as opposed to 'packing' them like we do now, since it's not entirely clear what the intention is, cfr. discussion we had with @stdweird

boegel:
@Caylo Can you look into cleaning up the `out.txt` that slipped in here, and write a test to verify the balancing when `--universe` is used?

boegel:
merged via #128

Issue #117, opened 2017-03-29
ipl_pin_list_direct syntax error when using partial nodes with local scheduler


boegel:
A warning like this appears when using the `local` scheduler on a partial node (e.g. ppn=16 on HPC-UGent's swalot, which has 20 cores per workernode):

```
IPL WARN> ipl_pin_list_direct syntax error, 16 list member should be -1, single CPU number, or CPU number range
```

For example (also occurs with `intel/2017a`):

```
$ module load intel/2016a
$ mympirun --sched=local hostname
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
IPL WARN> ipl_pin_list_direct syntax error, 16 list member should be -1, single CPU number, or CPU number range

$ echo $PBS_NUM_PPN
16

$ hostname
node2662.swalot.os
```

Does no occur with `pbs` scheduler:

```
$ mympirun --sched=pbs hostname
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2662.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
node2668.swalot.os
```

boegel:
@Caylo debug output from `mympirun` for a situation in which the warning occurs is available at https://gist.github.com/boegel/890002ca94e18747401a0a1fd4320571

it seems like the warning is only printed in some cases, i.e. it depends on which cores get allocated

log messages that look very relevant:

```
[...]

PR #122, opened 2017-04-26
dont set i_mpi_pin_processor_list

Caylo:
wrt #117 

still running on swalot, but the taskset output on delcatty looks good when cores are not just the first N cores (no warning):
```
pid 91230's current affinity list: 11
pid 91231's current affinity list: 12
pid 91232's current affinity list: 1
pid 91233's current affinity list: 4
```


boegel:
@Caylo did you verify that all `mympirun` tests we have in `vsc-testing` still pass with this change? (just wondering, not expecting any problems there)

Caylo:
@kehoste i didn't until after you asked, but everything still passes :)

boegel:
Going in, thanks @Caylo!

PR #126, opened 2017-05-17
get rid of perhost option

### LINKED ISSUE #32: Use machine file instead of -perhost
When you use PBS and the number of cores is not equally distributed over the hosts, mympirun makes wrong assumptions.

I think it would be cleaner to drop -perhost and use a machinefile with the correct distribution:

```
host1:n1
host2:n2
```

Caylo:
cleanup: we've been using the machinefile for this 
(also fixes #32)


boegel:
Did you rerun the `003_options.sh` test script in `vsc-testing` on top of this?

If it still passes, this is good to go...

Caylo:
@boegel yup, still works

PR #130, opened 2017-07-03
mympirun v4.0.0

boegel:


boegel:
@stdweird User interface is fully backwards compatible, so this should be a drop-in replacement.

All tests in `vsc-testing/mympirun` succeed with current `develop` branch of `vsc-mympirun`, and the `4.0.0-beta` installation didn't reveal any major problems (and the minor ones that were reported are solved), so let's roll it out?

stdweird:
sure, go ahead

Issue #136, opened 2017-10-05
--launcher ssh is not honored


boegel:
By default, `mympirun` uses `pbsdsh` as launcher without checking whether it's actually available.

If `mympirun` is told to use `ssh` instead via `--launcher ssh`, it actually tries to use the `pbsssh` wrapper, which again relies on `pbsdsh`.

Instead, it should do as it was told, and use plain `ssh` rather than sneakily trying to use `pbsdsh` instead.

PR #137, opened 2018-03-06
check whether specified launcher & launcher exec are available + fix faulty tests

boegel:
fix for #136

@stdweird I would like to have this solved before I look into SLURM support.

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -931,17 +931,7 @@ class MPI(object):
             else:
                 launcher_exec = self.get_rsh()
 
-            launcher_exec_opt = '-%s-exec %s' % (self.HYDRA_LAUNCHER_NAME, launcher_exec)
-            if which(launcher_exec):
-                # only specify launcher exec if specified command is available;
-                # only use pbsssh as launcher exec if pbsdsh is also available
-                if launcher_exec == 'pbsssh' and which('pbsdsh') is None:
-                    self.log.warning("Not using 'pbsssh' since 'pbsdsh' is not available in $PATH")
-                else:
-                    self.mpiexec_options.append(launcher_exec_opt)
-            else:
-                self.log.warning("Specified launcher exec '%s' is not available, so not adding '%s' option",
-                                 launcher_exec, launcher_exec_opt)
+            self.mpiexec_options.append("-%s-exec %s" % (self.HYDRA_LAUNCHER_NAME, launcher_exec))
 
     def get_hydra_info(self):
         """Get a dict with hydra info."""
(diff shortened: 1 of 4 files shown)


lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -927,8 +931,17 @@ def make_mpiexec_hydra_options(self):
             else:
                 launcher_exec = self.get_rsh()
 
-            self.mpiexec_options.append("-%s-exec %s" % (self.HYDRA_LAUNCHER_NAME, launcher_exec))
-
+            launcher_exec_opt = '-%s-exec %s' % (self.HYDRA_LAUNCHER_NAME, launcher_exec)
+            if which(launcher_exec):
+                # only specify launcher exec if specified command is available;
+                # only use pbsssh as launcher exec if pbsdsh is also available
+                if launcher_exec == 'pbsssh' and which('pbsdsh') is None:
```
stdweird: this is logic for `pbs.rm`

lib/vsc/mympirun/rm/pbs.py
```diff
@@ -37,13 +38,22 @@ class PBS(Sched):
...
-    HYDRA_LAUNCHER_EXEC = 'pbsssh'
     HYDRA_RMK = ['pbs']
 
+    def __init__(self, *args, **kwargs):
+        """PBS constructor"""
+        super(PBS, self).__init__(*args, **kwargs)
+
+        # pbsssh is only used if native support for pbsdsh is not supported in the MPI library;
+        # e.g. Intel MPI v5.x and newer supports using pbsdsh as launcher natively, no need for pbsssh wrapper
+        if which('pbsssh') and which('pbsdsh'):
```
stdweird: use a constant or the same string

lib/vsc/mympirun/rm/pbs.py
```diff
@@ -37,13 +38,22 @@ class PBS(Sched):
...
+            self.log.debug("Both 'pbsssh' and 'pbsdsh' found, so use 'pbsssh' as remote shell command.")
+            self.RSH_LARGE_CMD = 'pbsssh'
+            self.RSH_LARGE_LIMIT = 'pbsssh'
+            self.HYDRA_LAUNCHER_EXEC = 'pbsssh'
+        else:
+            self.log.debug("Either 'pbsssh' or 'pbsdsh' not available, not using 'pbsssh' as remote shell command.")
```
stdweird: can you also add what is being used to the message?

test/end2end.py
```diff
@@ -320,11 +328,10 @@ def test_env_variables(self):
             self.assertTrue(key in out, "%s is not in out" % key)
 
         regex = r'.*-envlist [^ ]*USER.*'
-        self.assertTrue(regex.find(out), "Variablesprefix USER isn't passed to mympirun script env")
```
stdweird: ha, that's a nice one. how did this one ever pass?

boegel:
```
>>> 'foo'.find('bar')
-1
>>> bool('foo'.find('bar'))
True
```

stdweird: `-1` should have guessed that.

PR #139, opened 2018-03-07
add support for --dry-run/-D option

boegel:
Example output:

```
$ mympirun --dry-run hostname
* output file            : None
* output timeout         : 3600
* fail on missing output : False
* mpirun command         : mpirun --file=/user/home/gent/vsc400/vsc40023/.mympirun_azji3b/SCHED_Coupler2018030716483995592_20180307_164839/mpdboot --machinefile /user/home/gent/vsc400/vsc40023/.mympirun_azji3b/SCHED_Coupler2018030716483995592_20180307_164839/nodes -genv MKL_NUM_THREADS '1' -genv I_MPI_PIN '1' -genv I_MPI_DAT_LIBRARY 'libdat2.so' -genv I_MPI_FABRICS 'shm:dapl' -genv LOADEDMODULES 'cluster/delcatty:vsc-base/2.5.8:vsc-mympirun/dry_run:GCCcore/6.4.0:binutils/2.28-GCCcore-6.4.0:icc/2018.1.163-GCC-6.4.0-2.28:ifort/2018.1.163-GCC-6.4.0-2.28:iccifort/2018.1.163-GCC-6.4.0-2.28:impi/2018.1.163-iccifort-2018.1.163-GCC-6.4.0-2.28:iimpi/2018.01:imkl/2018.1.163-iimpi-2018.01:intel/2018.01' -genv I_MPI_FALLBACK_DEVICE '0' -genv I_MPI_FALLBACK 'disable' -genv I_MPI_NETMASK '10.143.0.0/255.255.0.0' -genv I_MPI_DAPL_SCALABLE_PROGRESS '0' -genv MODULESHOME '/usr/share/lmod/lmod' -genv MODULEPATH '/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/sandybridge/modules/all:/etc/modulefiles/vsc' -np 16 -envlist LD_LIBRARY_PATH,PATH,PYTHONPATH,I_MPI_MPD_TMPDIR,I_MPI_ROOT,OMP_NUM_THREADS,MKL_EXAMPLES hostname
```

(still WIP because needs version bump + enhanced tests)

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -430,17 +430,17 @@ class MPI(object):
 
         run_kwargs = {
             'fatal_no_output': self.options.output_check_fatal,
-            'filename': None,
             'output_timeout': self.options.output_check_timeout,
         }
         if self.options.output:
             run_mpirun_cmd = RunFileLoopMPI.run
-            run_kwargs.update({'filename': self.options.output})
+            run_kwargs['filename'] = self.options.output
         else:
             run_mpirun_cmd = RunAsyncMPI.run
 
         if self.options.dry_run:
             self.log.info("Dry run, only printing generated mpirun command...")
+            run_kwargs['filename'] = None
             dry_run_output = '\n'.join([
                 "* output file            : %(filename)s",
                 "* output timeout         : %(output_timeout)s",
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -430,6 +430,7 @@ class MPI(object):
 
         run_kwargs = {
             'fatal_no_output': self.options.output_check_fatal,
+            'filename': None,
             'output_timeout': self.options.output_check_timeout,
         }
         if self.options.output:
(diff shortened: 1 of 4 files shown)


boegel:
@stdweird Please review

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -434,10 +434,26 @@ def main(self):
...
-        exitcode, _ = run_mpirun_cmd(self.mpirun_cmd, **run_kwargs)
+
+        if self.options.dry_run:
+            self.log.info("Dry run, only printing generated mpirun command...")
+
+            if 'filename' not in run_kwargs:
+                run_kwargs['filename'] = None
+
+            dry_run_output = '\n'.join([
+                "* output file            : %(filename)s",
```
stdweird: only print if output file is selected?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -434,10 +434,26 @@ def main(self):
...
+                "* output timeout         : %(output_timeout)s",
```
stdweird: why print this?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -434,10 +434,26 @@ def main(self):
...
+                "* fail on missing output : %(fatal_no_output)s",
```
stdweird: only when this is fatal

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -434,10 +434,26 @@ def main(self):
...
         if self.options.output:
             run_mpirun_cmd = RunFileLoopMPI.run
-            run_kwargs.update({'filename': self.options.output})
+            run_kwargs['filename'] = self.options.output
         else:
             run_mpirun_cmd = RunAsyncMPI.run
-        exitcode, _ = run_mpirun_cmd(self.mpirun_cmd, **run_kwargs)
+
+        if self.options.dry_run:
+            self.log.info("Dry run, only printing generated mpirun command...")
```
stdweird: move this to a function/method, so you can unittest it properly (instead of end2end)

stdweird:
i don't understand why you would print anything except the mpirun commandline. it seems useless

boegel:
@stdweird Simplified. I prefer testing end2end, since that's the only way to check whether the `--dry-run` option actually works (and to make sure no other output is generated).

stdweird:
what do you mean with "no other output"? you already add a logging.info, and i'm not sure you can control where that text ends up wrt the print statement
the end2end tests have very little value, typically they indicate there's something in the main that can't be mocked or should be run from it's own method/function.
it's not really different here imho. i understand that you would run one case and capture the output to make sure eg default logging is not info, but there's no added value in the extra opts or test both long and short optionname in the end2end; that behaviour is ideally tested in it's own unittests.

anyhow...

PR #140, opened 2018-03-12
add support for using mympirun on top of SLURM

boegel:
I want to clean up the new code in `rm/slurm.py` a bit so I can unit test it cleanly, especially the bit that parses `$SLURM_TASKS_PER_NODE`, hence `WIP`.


boegel:
@stdweird Ready for review, still needs some more testing in a SLURM environment.

I propose we go forward with leveraging `$SLURM_NODELIST` & `$SLURM_TASKS_PER_NODE` to determine the full list of available nodes (equivalent with `$PBS_NODEFILE`), we can improve on that in later revisions?

PR #145, opened 2018-06-06
[bug fix] use CmdList to construct mpirun command

boegel:
`mympirun` is utterly broken since v4.1.2 thanks to the naive switch to `run` in #144 , because the `mpirun` command isn't being constructed carefully enough.

This leads to errors like:

```
[mpiexec@node2128.delcatty.os] match_arg (../../utils/args/args.c:254): unrecognized argument machinefile /user/home/gent/vsc400/vsc40023/.mympirun_bfxlm3/4473084.master15.delcatty.gent.vsc_20180606_151207/nodes
[mpiexec@node2128.delcatty.os] HYDU_parse_array (../../utils/args/args.c:269): argument matching returned error
[mpiexec@node2128.delcatty.os] parse_args (../../ui/mpich/utils.c:4770): error parsing input array
[mpiexec@node2128.delcatty.os] HYD_uii_mpx_get_parameters (../../ui/mpich/utils.c:5106): unable to parse user arguments

Usage: ./mpiexec [global opts] [exec1 local opts] : [exec2 local opts] : ...
```

Each argument/option *must* be a separate item in the `self.mpirun_cmd` list.

The `CmdList` class added to `vsc-base` in https://github.com/hpcugent/vsc-base/pull/270 that wraps around `list` is a useful way to avoid these problems.

Tested with:

* full-cluster GROMACS on `delcatty` & `skitty`
* ...

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -38,7 +38,6 @@ import shutil
 import socket
 import stat
 import string
-import subprocess
 import sys
 import time
 
@@ -46,7 +45,7 @@ from distutils.version import LooseVersion
 from IPy import IP
 from vsc.utils.fancylogger import getLogger
 from vsc.utils.missing import get_subclasses, nub
-from vsc.utils.run import RunNoShell, RunAsyncLoopStdout, RunFile, RunLoop, run
+from vsc.utils.run import CmdList, RunNoShell, RunAsyncLoopStdout, RunFile, RunLoop, run
 
 # part of the directory that contains the installed fakes
 INSTALLATION_SUBDIRECTORY_NAME = '(VSC-tools|(?:vsc-)?mympirun)'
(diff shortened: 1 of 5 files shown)


lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -800,36 +800,36 @@ def get_localhosts(self):
     def make_mpdboot_options(self):
         """Add various options to mpdboot_options"""
 
-        self.mpdboot_options = self.MPDBOOT_OPTIONS[:]
+        self.mpdboot_options = CmdList(self.MPDBOOT_OPTIONS)
```
JensTimmerman: I think that copy of self.MPDBOOT_OPTIONS was there for a reason, better keep this a copy?

boegel: The copy is there to avoid that `self.MPDBOOT_OPTIONS` gets changed in-place when we add stuff to `self.mpbdoot_options`, but that's no longer required when using `CmdList` (the copying is done implicitly there).

boegel:
@stdweird please review

Issue #149, opened 2018-09-27
sm no longer support in OpenMPI 3.0.0


boegel:
```
As of version 3.0.0, the "sm" BTL is no longer available in Open MPI.

Efficient, high-speed same-node shared memory communication support in
Open MPI is available in the "vader" BTL.  To use the vader BTL, you
can re-run your job with:

    mpirun --mca btl vader,self,... your_mpi_application
```

boegel:
fixed in #150

PR #150, opened 2018-09-27
use 'vader' rather than 'sm' for OpenMPI 3.0 & newer

boegel:
fix for #149


boegel:
@stdweird Let me test a bit more this first before merging (assuming Jenkins approves it...)

PR #152, opened 2019-05-20
Add scatter to hybrid and make explicit

wpoely86:


test/00-import.py
```diff
@@ -1,4 +1,28 @@
 #
+# Copyright 2019-2019 Ghent University
```
stdweird: add a (empty) docstring here, so the header is not duplicated

test/__init__.py
```diff
@@ -1,4 +1,28 @@
 #
+# Copyright 2019-2019 Ghent University
```
stdweird: same docstring remark

lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -173,7 +176,7 @@ def pinning_override(self):
         if self.pinning_override_type in ('packed', 'compact', 'bunch'):
             cmd.add(['-env', 'I_MPI_PIN_PROCESSOR_LIST=allcores:map=bunch'])
         elif self.pinning_override_type in ('spread', 'scatter'):
-            cmd.add(['-env', 'I_MPI_PIN_PROCESSOR_LIST=allcores'])
+            cmd.add(['-env', 'I_MPI_PIN_PROCESSOR_LIST=allcores:map=%s' % self.pinning_override_type])
```
stdweird: what are the possible values?

stdweird:
@wpoely86 version bump?

lib/vsc/mympirun/mpi/intelmpi.py
wpoely86: https://software.intel.com/en-us/mpi-developer-reference-windows-environment-variables-for-process-pinning

bunch, spread and scatter

wpoely86:
The version bump is well hidden between all the style stuff :wink:

lib/vsc/mympirun/mpi/intelmpi.py
boegel: tests?

PR #153, opened 2019-05-22
Add tests for pinning

wpoely86:


wpoely86:
@boegel please review

boegel:
@wpoely86 I guess it makes sense to also test the default (i.e. without touching `pinning_override_type`)?

wpoely86:
The function is not called unless asked. 

```python
        if self.pinning_override_type is not None:
            self.mpirun_cmd.add(self.pinning_override())
```

boegel:
Nevermind, `pinning_override()` is never called when `pinning_override_type` is the default `None`, so good to go...

PR #154, opened 2019-05-22
specify that long description is in MarkDown format

boegel:
Uploading latest version to PyPI failed with:

```
INFO: running upload
Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

I hope this fixes that problem...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,10 +46,11 @@ PACKAGE = {
         'vsc-install >= 0.10.25',  # for modified subclassing
         'IPy',
     ],
-    'version': '4.1.8',
+    'version': '4.1.9',
     'author': [sdw, kh],
     'maintainer': [sdw, kh],
     'zip_safe': False,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,10 +46,11 @@
         'vsc-install >= 0.10.25',  # for modified subclassing
         'IPy',
     ],
-    'version': '4.1.8',
+    'version': '4.1.9',
     'author': [sdw, kh],
     'maintainer': [sdw, kh],
     'zip_safe': False,
+    'long_description_content_type': 'text/markdown',
```
stdweird: move it to vsc-install?

stdweird: adn there you can autodetect etc etc

boegel: general fix in https://github.com/hpcugent/vsc-install/pull/107

PR #162, opened 2019-12-18
run tests with both Python 2.7 & 3.6 via tox

boegel:


boegel:
@stdweird Any idea what's going on here?

Why would `vsc-install` try to pull in `vsc-base` from `github.ugent.be` (and why doesn't it ignore that failure and try `github.com` instead)?

boegel:
```
  ERROR: Command errored out with exit status 128:
   command: git clone -q 'ssh://****@github.ugent.be/hpcugent/vsc-base.git' /tmp/pip-wheel-1IGUwM/vsc-base
       cwd: None
  Complete output (5 lines):
  ERROR: Repository not found.
  fatal: Could not read from remote repository.
  
  Please make sure you have the correct access rights
  and the repository exists.
  ----------------------------------------
ERROR: Command errored out with exit status 128: git clone -q 'ssh://****@github.ugent.be/hpcugent/vsc-base.git' /tmp/pip-wheel-1IGUwM/vsc-base Check the logs for full command output.
```

stdweird:
@boegel it thinks it (or one of its dependencies) is a private repo. notmally it fails and tries the next from a set of url, maybe seomthing in setuptools is more strict these days?

stdweird:
git ssh is used for cloning, vsc-install sees that as a sign of a private repo (why else would you need ssh). question is, why is ssh used (and why did this work before). is the jenkins config for mympirun different somehow?

boegel:
@stdweird I don't see a difference with `vsc-config` (which also requires `vsc-base`)...

boegel:
@stdweird `vsc-config` was also cloned via `git@`, and there the issue doesn't occur:

```
jenkins@jenkins8:~/jobs/hpc ugent github.com/jobs/vsc-config/branches/PR-98/workspace$ git remote -v
origin	git@github.com:hpcugent/vsc-config.git (fetch)
origin	git@github.com:hpcugent/vsc-config.git (push)
```

boegel:
This is from the last test on `master` for `vsc-mympirun`:

```
...
INFO: Searching for vsc-base>=2.9.3
INFO: Doing git clone from ssh://git@github.ugent.be/hpcugent/vsc-base.git to /tmp/easy_install-hQsEPK/vsc-base.git
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
INFO: Doing git clone from ssh://git@github.com/hpcugent/vsc-base.git to /tmp/easy_install-hQsEPK/vsc-base.git
INFO: Best match: vsc-base 2.9.3
...
```

So there the failing clone from `github.ugent.be` was also tried, but not fatal...

boegel:
The issue could be that `tox` uses `pip` by default to install dependencies, see https://tox.readthedocs.io/en/latest/example/basic.html#further-customizing-installation .

But that doesn't explain why it doesn't happen with `vsc-config`...

wpoely86:
We've seen this happen before. The cloneing fails to continue after github.ugent.be doesn't work. It never happened for me but on @smoors his laptop it happened regularly. I haven't found time to debug it but as far as I remember your Jenkins runs Debian and @smoors laptop was Ubuntu. So it might be more related to that then to the code changes.

boegel:
Looks like setuptools recently basically replaced easy_install with pip, see https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0, that could have something to do with it...

stdweird:
it probably is due to pip. we need https://github.com/hpcugent/vsc-install/issues/121 to be sure
the release notes state that pip does not work with eggs, and vsc-base and vsc-config are zip safe (ie egg is used); but mympirun isn't (no egg, so pip is picked up)

boegel:
Issues should hopefully be fixed with https://github.com/hpcugent/vsc-install/pull/122...

PR #166, opened 2020-03-29
port to Python 3

boegel:
This should make all tests pass with both Python 2 and 3...

### DIFF
--- a/lib/vsc/mympirun/common.py
+++ b/lib/vsc/mympirun/common.py
@@ -96,8 +96,18 @@ def what_sched(requested, schedm):
 
     # next, try to use the scheduler defined by environment variables
     for sched in found_sched:
-        nodeinfo = not hasattr(sched, 'SCHED_ENVIRON_NODE_INFO') or sched.SCHED_ENVIRON_NODE_INFO in os.environ
-        if nodeinfo and sched.SCHED_ENVIRON_ID in os.environ:
+
+        # take into account that SCHED_ENVIRON_NODE_INFO can be None,
+        # and chechking "None in os.environ" fails hard in Python 3 (string value is required)
+        if hasattr(sched, 'SCHED_ENVIRON_NODE_INFO'):
+            if sched.SCHED_ENVIRON_NODE_INFO is not None:
+                nodeinfo = sched.SCHED_ENVIRON_NODE_INFO in os.environ
+            else:
+                nodeinfo = False
+        else:
+            nodeinfo = True
+
+        if nodeinfo and sched.SCHED_ENVIRON_ID is not None and sched.SCHED_ENVIRON_ID in os.environ:
             return sched, found_sched
 
     # If that fails, try to force the local scheduler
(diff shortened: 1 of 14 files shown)


lib/vsc/mympirun/common.py
```diff
@@ -96,8 +96,18 @@ def sched_to_key(klass):
 
     # next, try to use the scheduler defined by environment variables
     for sched in found_sched:
-        nodeinfo = not hasattr(sched, 'SCHED_ENVIRON_NODE_INFO') or sched.SCHED_ENVIRON_NODE_INFO in os.environ
-        if nodeinfo and sched.SCHED_ENVIRON_ID in os.environ:
+
+        # take into account that SCHED_ENVIRON_NODE_INFO can be None,
+        # and chechking "None in os.environ" fails hard in Python 3 (string value is required)
```
stdweird, line 101: chechking -> checking

lib/vsc/mympirun/common.py
```diff
@@ -96,8 +96,18 @@ def sched_to_key(klass):
...
+        if hasattr(sched, 'SCHED_ENVIRON_NODE_INFO'):
+            if sched.SCHED_ENVIRON_NODE_INFO is not None:
+                nodeinfo = sched.SCHED_ENVIRON_NODE_INFO in os.environ
+            else:
+                nodeinfo = False
+        else:
+            nodeinfo = True
+
+        if nodeinfo and sched.SCHED_ENVIRON_ID is not None and sched.SCHED_ENVIRON_ID in os.environ:
```
stdweird, line 110: if nodeinfo is true, the attr might not exist (last else block)
the `sched.SCHED_ENVIRON_ID is not None and sched.SCHED_ENVIRON_ID in os.environ` was already tested before, not sure why it is tested again

lib/vsc/mympirun/common.py
```diff
@@ -96,8 +96,18 @@ def sched_to_key(klass):
 
     # next, try to use the scheduler defined by environment variables
     for sched in found_sched:
-        nodeinfo = not hasattr(sched, 'SCHED_ENVIRON_NODE_INFO') or sched.SCHED_ENVIRON_NODE_INFO in os.environ
-        if nodeinfo and sched.SCHED_ENVIRON_ID in os.environ:
+
+        # take into account that SCHED_ENVIRON_NODE_INFO can be None,
+        # and chechking "None in os.environ" fails hard in Python 3 (string value is required)
+        if hasattr(sched, 'SCHED_ENVIRON_NODE_INFO'):
```
stdweird, line 102: should be 
```
nodeinfo = getattr(sched, 'SCHED_ENVIRON_NODE_INFO', True) and (sched.SCHED_ENVIRON_NODE_INFO or '') in os.environ
```

lib/vsc/mympirun/common.py
boegel: I think you're mixing up `SCHED_ENVIRON_NODE_INFO` and `SCHED_ENVIRON_ID`?

lib/vsc/mympirun/common.py
boegel: This is not correct, tests fail now, and it makes sense: if `SCHED_ENVIRON_NODE_INFO` is not an attribute, then you get `True and ...` where the `...` does a `sched.SCHED_ENVIRON_NODE_INFO` which fails hard.

PR #171, opened 2020-10-22
use UCX as PML with OpenMPI 4.x and newer, if ompi_info reports it as supported

boegel:
The OpenMPI 4.0.3 we have as a part of `foss/2020a` no longer supports `openib` as BTL, because it was built with support for UCX (see https://github.com/easybuilders/easybuild-easyblocks/pull/2188).

The changes in `vsc/mympirun/mpi/openmpi.py` make sure that `--mca pml ucx` is used with OpenMPI 4.x if `ompi_info` reports `ucx` as supported PML.

The other changes are trivial style cleanups.

~~WIP, because the tests should be enhanced to cover this change~~

### DIFF
--- a/lib/vsc/mympirun/mpi/openmpi.py
+++ b/lib/vsc/mympirun/mpi/openmpi.py
@@ -232,7 +232,7 @@ class OpenMpi3(OpenMpiOversubscribe):
     }
 
 
-class OpenMpi4(OpenMpiOversubscribe):
+class OpenMpi4(OpenMpi3):
 
     """
     An implementation of the MPI class for OpenMPI 4.x & more recent.
--- a/lib/vsc/mympirun/mpi/openmpi.py
+++ b/lib/vsc/mympirun/mpi/openmpi.py
@@ -243,7 +243,11 @@ class OpenMpi4(OpenMpiOversubscribe):
     def use_ucx_pml(self):
         """Determine whether or not to use the UCX Point-to-Point Messaging Layer (PML)."""
 
-        # if 'ompi_info' reports that UCX is a supported PML
+        # use UCX if 'ompi_info' reports that it is a supported PML;
+        # only for OpenMPI 4.x and newer, as recommended by UCX:
+        # "OpenMPI supports UCX starting from version 3.0, but it's recommended to use
+        # "version 4.0 or higher due to stability and performance improvements."
+        # (see https://openucx.github.io/ucx/running.html#openmpi-with-ucx)
         cmd = "ompi_info"
         ec, out = run(cmd)
         if ec:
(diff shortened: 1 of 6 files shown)


lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -219,3 +230,24 @@ class OpenMpi3(OpenMpiOversubscribe):
...
+class OpenMpi4(OpenMpiOversubscribe):
+
+    """
+    An implementation of the MPI class for OpenMPI 4.x & more recent.
+    """
+
+    _mpirun_version = staticmethod(lambda ver: version_in_range(ver, '4', None))
+
+    def use_ucx_pml(self):
+        """Determine whether or not to use the UCX Point-to-Point Messaging Layer (PML)."""
```
stdweird, line 244: why is this liited to openmpi 4? any openmpi with ucx support should probably use

boegel: Several reasons:

* OpenMPI 4 (used in `2020a` toolchains) is a natural cutoff, since it's a new major version (`2019b` and older use OpenMPI <=3)
* "OpenMPI supports UCX starting from version 3.0, but it’s recommended to use version 4.0 or higher due to stability and performance improvements." (see https://openucx.github.io/ucx/running.html)
* The OpenMPI 3.1.4 modules we have in `foss/2019b` have UCX support, but rely on the `ucx` RPM we have installed in the OS (while in `foss/2020a` we install `UCX` as a proper dep for OpenMPI 4.0.3, so we're in control)
* We've been using `openib` rather than `ucx` for a while now in `foss/2019b`, which mostly works fine, so why change that...
* `mympirun` is currently broken with `foss/2020a`, because the OpenMPI in there no longer supports `openib` (see https://github.com/easybuilders/easybuild-easyblocks/pull/2188), this PR fixes that by using `ucx` instead

We can still consider also using the `ucx` PML with OpenMPI 3.x, but then we need to reinstall the OpenMPI 3 modules with a proper UCX dep (so we're in control), and do more thorough testing.

stdweird: i thought we couldn't build the `openib` due to missing headers, and that it was thus also not avail on openmpi3.
can you also add the 2nd item in a comment or in docstring?

boegel: OpenMPI 3 was being configured with `--without-verbs` on RHEL8 because of a bug in EasyBuild which is now fixed (see https://github.com/easybuilders/easybuild-framework/pull/3477), but we still need to reinstall OpenMPI 3 on `doduo`, that's on our TODO list before we give pilot users access.

stdweird (approved):

PR #172, opened 2020-10-26
fix use of mypmirun with OpenMPI 3.1.x (covers foss/2018b to foss/2019b)

boegel:


boegel:
@stdweird With this version I can run MPI hello world with `mypmirun` on `doduo` with `foss/2019b`, `foss/2020a`, `intel/2019b` and `intel/2020a`

stdweird (approved):

PR #173, opened 2020-10-26
mympirun fixes for recent Intel MPI versions

### LINKED ISSUE #169: I_MPI_MPD_TMPDIR environment variable is not supported
with the latest version, 5.1.0, I get the following message upon startup:

"""
> [0] MPI startup(): I_MPI_MPD_TMPDIR environment variable is not supported.
> [0] MPI startup(): Similar variables:
>          I_MPI_TMPDIR
> [0] MPI startup(): To check the list of supported variables, use the impi_info utility or refer to https://software.intel.com/en-us/mpi-library/documentation/get-started.
"""

after that, the test program runs fine.

full debug output: https://gist.github.com/smoors/f11b317dc2d3ae2ca1f980dd68ba3668

boegel:
* also pass down `$FI_PROVIDER*` environment variables (fixes crash when using `mympirun` with `intel/2020a` or other toolchains that include `impi` 2019.x)
* define `$I_MPI_TMPDIR` rather than `$I_MPI_MPD_TMPDIR` (>= 2019) (fixes #169)

WIP because may require fixes in tests + version bump is missing (awaiting merge of #172)

### DIFF
--- a/lib/vsc/mympirun/mpi/intelmpi.py
+++ b/lib/vsc/mympirun/mpi/intelmpi.py
@@ -44,6 +44,7 @@ def _enable_disable(boolvalue):
     """Return enable/disable for boolean value"""
     return {True: 'enable', False: 'disable'}.get(bool(boolvalue))
 
+
 def _one_zero(boolvalue):
     """Return enable/disable for boolean value"""
     return int(bool(boolvalue))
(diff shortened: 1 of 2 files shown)


lib/vsc/mympirun/mpi/intelmpi.py
```diff
@@ -97,15 +98,19 @@ def make_mpdboot_options(self):
             mpd = which(['mpd.py'])
             self.mpdboot_options.append('-m "\\\"%s --bulletproof\\\""' % mpd)
 
+    def set_impi_tmpdir(self):
+        """Set location of temporary directory that Intel MPI should use."""
+        # this one also needs to be set at runtime
+        impi_tmpdir = tempfile.gettempdir()
```
stdweird, line 104: i'm going to merge this, but it is not a garantee that this is unique across all nodes

stdweird:
@boegel no version bump?

stdweird (approved):

lib/vsc/mympirun/mpi/intelmpi.py
boegel: I think Intel MPI takes care of that by itself, it just wants a place to use as a scratch location?

PR #184, opened 2021-11-22
use --map-by core --bind-to core by default with OpenMPI (WIP)

boegel:
to ensure sequential ranking/pinning of MPI processes by cores (rather than by NUMA domain)

Still WIP, since I want to catch this in the tests too (actually, the fact that the tests still pass while making a change like this is bad imho)

### DIFF
--- a/lib/vsc/mympirun/mpi/openmpi.py
+++ b/lib/vsc/mympirun/mpi/openmpi.py
@@ -85,10 +85,16 @@ class OpenMPI(MPI):
         """Set mpiexec options"""
         super(OpenMPI, self).set_mpiexec_options()
 
-        # specify number of processes to start per node if --hybrid is used
         if self.options.hybrid:
+            # specify number of processes to start per node if --hybrid is used
             procs_per_node = self.multiplier * self.options.hybrid
             self.mpiexec_options.add(['--map-by', 'ppr:%s:node' % procs_per_node])
+        else:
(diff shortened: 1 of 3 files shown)


lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -85,10 +85,16 @@ def set_mpiexec_options(self):
...
         if self.options.hybrid:
+            # specify number of processes to start per node if --hybrid is used
             procs_per_node = self.multiplier * self.options.hybrid
             self.mpiexec_options.add(['--map-by', 'ppr:%s:node' % procs_per_node])
+        else:
+            # map MPI processes by core (default is --map-by numa),
+            # and bind to core (default is --bind-to numa' when # ranks > 2),
+            # to match default behaviour of Intel MPI;
+            # this is important for performance for OpenFOAM for example, especially on AMD Rome CPUs
+            self.mpiexec_options.add(['--map-by', 'core', '--bind-to', 'core'])
```
stdweird, line 97: can't we also use `--map-by ppr:H:node` here? not sure it matters much for non-hybrid, but code might be simpler that way

boegel: `ppr:N:node` where `N` is the number of ranks (same as available core) basically means "start N processes per node", which is basically the default, I don't think it implies sequential ranking at all.

stdweird (approved):

PR #187, opened 2022-02-14
ensure proper core pinning of processes for OpenMPI by correcting value passed to --map-by 

stdweird:


test/end2end.py
```diff
@@ -508,7 +508,7 @@ def test_openmpi4(self):
         self.assertTrue(regex.search(out), "Pattern '%s' should be found in: %s" % (regex.pattern, out))
 
         # mapping/binding to core is done by default
-        regex = re.compile(r"^fake mpirun called with args:.*--map-by core --bind-to core")
+        regex = re.compile(r"^fake mpirun called with args:.*--map-by ppr:2:node:PE=1:SPAN:NOOVERSUBSCRIBE")
```
boegel, line 511: Why is this `ppr:2`? Will this test only pass on a dual-core system?

test/end2end.py
```diff
@@ -290,7 +290,7 @@ def test_option_hybrid_openmpi(self):
         """Test --hybrid command line option with OpenMPI"""
         install_fake_mpirun('mpirun', self.tmpdir, 'openmpi', '3.1')
         ec, out = run([sys.executable, self.mympiscript, '--hybrid', '5', 'hostname'])
-        self.assertTrue(" --map-by ppr:5:node " in out)
+        self.assertTrue(" --map-by ppr:5:node:PE=2469:SPAN:NOOVERSUBSCRIBE " in out)
```
boegel, line 293: Where does `PE=2469` come from?!

test/end2end.py
boegel: Can we also add a test case where oversubscription is being allowed?

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -85,16 +99,26 @@ def set_mpiexec_options(self):
...
-            # and bind to core (default is --bind-to numa' when # ranks > 2),
-            # to match default behaviour of Intel MPI;
-            # this is important for performance for OpenFOAM for example, especially on AMD Rome CPUs
-            self.mpiexec_options.add(['--map-by', 'core', '--bind-to', 'core'])
+            over = 'NO'
+
+        # make sure we start enough per node so it can fill the total_number_of_processes
+        tnop = self.total_number_of_processes()
+        un = len(self.nodes_uniq)
+        ppn = int(math.ceil(tnop / un))
```
boegel, line 111: The variable names are a bit cryptic here, how about:

* `tnop` -> `tot_procs`
* `un` -> `uniq_nodes`
* `ppn` -> `cores_per_node`

stdweird:
tested and verified with bunch of common hybird/multi and regular usage

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -85,16 +99,26 @@ def set_mpiexec_options(self):
...
+        logging.debug("Setting up map for %s (%s total number of processes on %s unique nodes)", ppn, tnop, un)
+
+        mapby = [
+            'ppr', "%s" % ppn, 'node',
```
boegel, line 115: Can we add some comments here to explain what each of these lines mean?
Maybe also include a pointer to https://www.open-mpi.org/doc/current/man1/mpirun.1.php on top of `mapby =`?

`ppr` vs `ppn` is very confusing here, but this line basically is:

```
# specify number of cores to start per node;
# 'ppr' is processes per resource type, with 'node' being the resource type
```

Also, use `str(cores_per_node)` rather than `"%s" % cores_per_node`?

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -85,16 +99,26 @@ def set_mpiexec_options(self):
...
+            "PE=%s" % os.environ['OMP_NUM_THREADS'],
```
boegel, line 116:
```
# 'PE=' specifies number of processing elements (cores) to assign to each MPI process
```

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -85,16 +99,26 @@ def set_mpiexec_options(self):
...
+            "SPAN",
```
boegel, line 117:
```
# 'SPAN' specifies to load balance the processes across the allocation

File: lib/vsc/mympirun/mpi/openmpi.py [hunk 7]
```diff
@@ -85,16 +99,26 @@ def set_mpiexec_options(self):
...
+            "%sOVERSUBSCRIBE" % over,
```
boegel, line 118:
```
# if more MPI processes are started than there are cores available,
# we need to indicate that's deliberate by specifying 'OVERSUBSCRIBE'
```

Maybe this line should become:

```python
'OVERSUBSCRIBE' if self.is_oversubscribed() else 'NOOVERSUBSCRIBE'

File: lib/vsc/mympirun/mpi/openmpi.py [hunk 8]
```diff
@@ -71,10 +72,23 @@ def set_mpiexec_global_options(self):
...
+            if self.options.stats:
+                # report UCX stats at the end to stdout
+                os.environ['UCX_STATS_TRIGGER'] = 'exit'
+                os.environ['UCX_STATS_DEST'] = 'stdout'
         else:
             # default PML (ob1) uses Byte Transport Layer (BTL)
             self.mpiexec_global_options['btl'] = self.device
 
+        if self.options.debuglvl > 3:
+            value = [None, "--%(name)s"]
```
boegel, line 88: Please add a comment here to explain that the `None` here is not a mistake, this looks really weird without knowing that this is going to be processed. And you should really use a tuple `(None, "--%(name)s")` here imho, a list is mutable which makes no sense at all here.

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -855,7 +864,13 @@ def get_mpiexec_global_options(self):
             else:
                 # insert the keyvalue pair into the correct command line argument
                 # the command for setting the environment variable depends on the mpi flavor
-                opts.add(self.MPIEXEC_TEMPLATE_GLOBAL_OPTION, tmpl_vals={'name': key, "value": val})
+                if isinstance(val, (list, tuple)):
```
boegel, line 867: Also check that length is exactly `2`?

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -855,7 +864,13 @@ def get_mpiexec_global_options(self):
...
+                    template = val[1]
+                    val = val[0]
```
boegel, line 869: `val, template = val`

boegel (changes requested):

test/end2end.py
stdweird: the ppn is 12345

lib/vsc/mympirun/mpi/openmpi.py
stdweird: i'm not repeating text from a manpage here.

test/end2end.py
boegel: Still not clear to me why this is `ppr:2` here, isn't this basically hardcoding to Jenkins which has 2 cores?

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -106,13 +107,15 @@ def set_mpiexec_options(self):
...
-        un = len(self.nodes_uniq)
-        ppn = int(math.ceil(tnop / un))
-        logging.debug("Setting up map for %s (%s total number of processes on %s unique nodes)", ppn, tnop, un)
+        tot_processes = self.total_number_of_processes()
+        unique_nodes = len(self.nodes_uniq)
+        processes_per_node = int(math.ceil(tot_processes / unique_nodes))
+        logging.debug("Setting up map for %s (%s total number of processes on %s unique nodes)",
+                      processes_per_node, tot_processes, unique_nodes)
 
+        # See https://www.open-mpi.org/doc/current/man1/mpirun.1.php "Mapping, Ranking, and Binding: Oh My!"
```
boegel, line 116: @stdweird Let's change this to a version-specific URL, because this stuff will probably change in OpenMPI v5:

```suggestion
        # See https://www.open-mpi.org/doc/v4.1/man1/mpirun.1.php "Mapping, Ranking, and Binding: Oh My!"

boegel (changes requested):

boegel (approved):

PR #189, opened 2022-03-08
set $OMP_PROC_BIND to 'true' wen using --hybrid to ensure proper binding of OpenMP threads

stdweird:


test/mpi.py
```diff
@@ -164,11 +164,27 @@ def test_is_mpirun_for(self):
 
     def test_set_omp_threads(self):
         """test if OMP_NUM_THREAD gets set correctly"""
+        del os.environ['OMP_NUM_THREADS']
+        del os.environ['OMP_PROC_BIND']
```
boegel, line 168: This should be done conditionally, because if they're not defined, it'll break the test?

boegel (changes requested):

test/mpi.py
stdweird: yep

stdweird: @boegel fixed

boegel (approved):

PR #197, opened 2024-03-26
disable libfabric (OFI) btl + mtl by default when using OpenMPI (HPC-11091)

boegel:
Motivation for this are the inconsistent errors "`Failed to modify UD QP to INIT on mlx5_0: Operation not permitted`" that we have been seeing after updating to OFED 23.10.

Worth noting, same can be achieved in contexts where `mympirun` is not used via:

```
export OMPI_MCA_btl='^uct,ofi'
export OMPI_MCA_pml='ucx'
export OMPI_MCA_mtl='^ofi'
```

### DIFF
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2011-2023 Ghent University
+# Copyright 2011-2024 Ghent University
 #
 # This file is part of vsc-mympirun,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -601,8 +601,7 @@ class MPI(MpiBase):
                              "(text file with minimal entry 'password=<somesecretpassword>')"), mpdconffn)
 
             with open(mpdconffn, 'w') as mpdconff:
-                mpdconff.write("password=" + ''.join(random.choice(string.ascii_uppercase + string.digits)
-                                                     for x in range(10)))
+                mpdconff.write("password=" + ''.join(random.choices(string.ascii_uppercase + string.digits, k=10)))
             # set correct permissions on this file.
             os.chmod(mpdconffn, stat.S_IREAD)
 
--- a/lib/vsc/mympirun/mpi/mpi.py
+++ b/lib/vsc/mympirun/mpi/mpi.py
@@ -601,8 +601,8 @@ class MPI(MpiBase):
                              "(text file with minimal entry 'password=<somesecretpassword>')"), mpdconffn)
 
             with open(mpdconffn, 'w') as mpdconff:
-                mpdconff.write("password=%s" % ''.join(random.choice(string.ascii_uppercase + string.digits)
-                                                       for x in range(10)))
+                mpdconff.write("password=" + ''.join(random.choice(string.ascii_uppercase + string.digits)
+                                                     for x in range(10)))
             # set correct permissions on this file.
             os.chmod(mpdconffn, stat.S_IREAD)
 
--- a/lib/vsc/mympirun/pmi/sched.py
+++ b/lib/vsc/mympirun/pmi/sched.py
@@ -53,9 +53,8 @@ class Info:
 
     def __str__(self):
         """Human readable"""
-        res = f"{self.nodes} nodes; with per node {self.cores} cores, {self.ranks} ranks, "
-        res += "{self.mem} mem, {self.gpus} gpus"
-        return res
+        return (f"{self.nodes} nodes; with per node {self.cores} cores, "
+            f"{self.ranks} ranks, {self.mem} mem, {self.gpus} gpus")
 
     def deepcopy(self):
         """Return a (deep)copy"""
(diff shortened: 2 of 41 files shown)


lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -601,8 +601,8 @@ def make_mpdboot(self):
                              "(text file with minimal entry 'password=<somesecretpassword>')"), mpdconffn)
 
             with open(mpdconffn, 'w') as mpdconff:
-                mpdconff.write("password=%s" % ''.join(random.choice(string.ascii_uppercase + string.digits)
-                                                       for x in range(10)))
+                mpdconff.write("password=" + ''.join(random.choice(string.ascii_uppercase + string.digits)
+                                                     for x in range(10)))
```
wdpypere, line 605: you don't need the range if you use `choices()`:
```
f"password={"".join(random.choices(string.ascii_uppercase + string.digits, k=10))}"
```

lib/vsc/mympirun/pmi/sched.py
```diff
@@ -53,8 +53,9 @@ def __init__(self, nodes=None, cores=None, ranks=None, mem=None, gpus=None):
 
     def __str__(self):
         """Human readable"""
-        return "{} nodes; with per node {} cores, {} ranks, {} mem, {} gpus".format(
-            self.nodes, self.cores, self.ranks, self.mem, self.gpus)
+        res = f"{self.nodes} nodes; with per node {self.cores} cores, {self.ranks} ranks, "
+        res += "{self.mem} mem, {self.gpus} gpus"
+        return res
```
wdpypere, line 58: you don't need to have a variable to split the lines. you can use parentheses to split a long string over more than 1 line. it will still print a single line:
```
return (f" my first part of {self.string}"
    "my second part")
```
for example:
```
>>> x='test'
>>> print((f"allo {x} more_test"
... "allo")
... )
allo test more_testallo
```

wdpypere:
LGTM, very minor remarks.

boegel:
Don't merge this yet please, we should actually test first that the workaround implemented here fixes the issues...

wdpypere:
> Don't merge this yet please, we should actually test first that the workaround implemented here fixes the issues...

ok

lib/vsc/mympirun/mpi/mpi.py
```diff
@@ -601,8 +601,7 @@ def make_mpdboot(self):
                              "(text file with minimal entry 'password=<somesecretpassword>')"), mpdconffn)
 
             with open(mpdconffn, 'w') as mpdconff:
-                mpdconff.write("password=%s" % ''.join(random.choice(string.ascii_uppercase + string.digits)
-                                                       for x in range(10)))
+                mpdconff.write("password=" + ''.join(random.choices(string.ascii_uppercase + string.digits, k=10)))
```
wdpypere, line 604: why u no f-string? :wink:

lib/vsc/mympirun/pmi/sched.py
```diff
@@ -53,8 +53,8 @@ def __init__(self, nodes=None, cores=None, ranks=None, mem=None, gpus=None):
 
     def __str__(self):
         """Human readable"""
-        return "{} nodes; with per node {} cores, {} ranks, {} mem, {} gpus".format(
-            self.nodes, self.cores, self.ranks, self.mem, self.gpus)
+        return (f"{self.nodes} nodes; with per node {self.cores} cores, "
+            "{self.ranks} ranks, {self.mem} mem, {self.gpus} gpus")
```
wdpypere, line 57: you will need and `f` prefixed on this line as well, every line is handled separately

lib/vsc/mympirun/mpi/mpi.py
boegel: I don't see the point here, it would only make the code harder to read by adding `{...}`

lib/vsc/mympirun/pmi/sched.py
boegel: Oh right, good catch, thanks, will fix

boegel: fixed in force push of b400be0

lib/vsc/mympirun/mpi/mpi.py
wdpypere: it's faster. 
And I find the f-string notation more readable but that's personal preference I guess.
not going to die on this hill though. :)

boegel:
@hajgato is testing this currently, seems to work as designed for me with a quick test (MPI hello on top of OpenMPI)

lib/vsc/mympirun/mpi/mpi.py
boegel: "faster" makes no sense here as an argument, this password file is only created if it doesn't exist yet, and it'll be done in microseconds using f-strings or not :)

wdpypere: sure, but the best way to learn new practices is to use them all the time. circumventing the linter check by doing the `+` is not good practice. :wink: but again, I won't die on this hill.

wdpypere:
> @hajgato is testing this currently, seems to work as designed for me with a quick test (MPI hello on top of OpenMPI)

so this can be merged?

hajgato:
@wdpypere  Yes, tests were passed on Tier1

PR #199, opened 2025-01-13
fix compatibility with OpenMPI 5.x

boegel:


Lewih:
I advise Changing the prrte option to 
`--mca prte_keep_fqdn_hostnames 1`

lib/vsc/mympirun/mpi/openmpi.py
```diff
@@ -306,3 +306,24 @@ def use_ucx_pml(self):
...
+        """Set mpiexec options"""
+        # don't set orte_keep_fqdn_hostnames, since ORTE is no longer used starting OpenMPI 5.0,
+        # see also https://docs.open-mpi.org/en/v5.0.x/launching-apps/pmix-and-prrte.html,
+        del self.mpiexec_global_options['orte_keep_fqdn_hostnames']
+
+        super().set_mpiexec_options()
+
+        # use keep_fqdn_hostname PRRTE MCA parameter instead: --prtemca keep_fqdn_hostnames 1;
+        # see also https://docs.open-mpi.org/en/v5.0.x/mca.html#label-running-setting-mca-param-values
+        self.mpiexec_options.add(['--prtemca', 'keep_fqdn_hostnames', '1'])
```
hajgato: the right option is:
`--prtemca prte_keep_fqdn_hostnames 1`

hajgato:
mpirun 5 uses srun to set up daemons (for example `prted`). Since we submit with `--export="NONE"`, slurm exports `SLURM_EXPORT_ENV=NONE`, so all subsequent srun will run with an empty env list, therefore prted will fail (as it is no path).
1) in mympirun we implement setting `env SLURM_EXPORT_ENV=ALL mpirun -x SLURM_EXPORT_ENV=<previous_value> ...` (this makes sure that mpi initialization will run fine, and in the MPI environment the user selected value will be seen., although I am not sure that users would spawn new slurmsteps from an mpi task)
2) we set in jobcli in the slurm script header `unset  SLURM_EXPORT_ENV` before the script body comes.

I prefer the 2nd option, although in the case of manually set SLURM_EXPORT_ENV=NIL or NONE mympirun will fail.

hajgato (approved):

backelj:
> This is bloody hard to find, so logging it here: correct environment variable with OpenMPI 5.x is `$PRTE_MCA_prte_keep_fqdn_hostnames`

@boegel for the sake of completeness, when using foss/2026.1 we had to set both
```
PRTE_MCA_prte_keep_fqdn_hostnames=1
PMIX_MCA_pmix_keep_fqdn_hostnames=1
```
to make it work again.

The second one was added in [PR 3654](https://github.com/openpmix/openpmix/pull/3654) and appears in the sources for PMIx 6.1.0.

